PR #2, opened 2015-06-01
libgdiplus.so.0.0.0 from libgdiplus_3.12-0xamarin1_amd64.deb

dlhartveld:
... so that "it" works again.

The current libgdiplus is not ok, apparently. :'-( But this new version is. :-)


jtwaleson:
We should incorporate the vagrant building of resources like this, for reference see https://github.com/cloudfoundry/staticfile-buildpack

PR #15, opened 2015-12-04
Update backup credentials from Buildpack

RaphaelAudet:
Update backup credentials from Buildpack

### DIFF
--- a/start.py
+++ b/start.py
@@ -6,8 +6,8 @@ import signal
 import subprocess
 import time
 import sys
-import requests
 sys.path.insert(0, 'lib')
+import requests
 import buildpackutil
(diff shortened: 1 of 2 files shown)


start.py
```diff
@@ -6,6 +6,7 @@
 import subprocess
 import time
 import sys
+import requests
```
pommi: This line should be after ```sys.path.insert(0, 'lib')```, because the requests library is provided in the lib folder.

pommi: Fixed in c4896e5

rcefala:
This is superseded by backup_apiKey_in_headers_rebased branch

PR #24, opened 2016-03-01
Nginx: expose /_mxadmin/ for ADMIN_PORT with HTTP BasicAuth

djvdorp:


djvdorp:
Tested and found working in

- cf-test
- PWS

PR #32, opened 2016-04-19
x-frame-options support

emirozer:


start.py
```diff
@@ -45,7 +70,7 @@ def set_up_nginx_files():
         'ADMIN_PORT', str(admin_port)
     ).replace(
         'ROOT', os.getcwd()
-    )
+    ).replace('XFRAMEOPTIONS', x_frame_options)
```
djvdorp: For consistency, we might want to also put this `.replace` on a new line as the ones above do?

start.py
```diff
@@ -21,6 +21,31 @@
...
 
+def x_frame_opt_headers_handler(x_frame_opt, x_frame_opt_uri):
+    base = "add_header X-Frame-Options "
+
+    if x_frame_opt_uri:
+        return base + "\"ALLOW-FROM " + x_frame_opt_uri + "\";"
+
+    if x_frame_opt:
+        if x_frame_opt == "Never allow":
+            return base + "\"DENY\";"
```
jtwaleson: better to mix ' and " so '"DENY"'

start.py
```diff
@@ -21,6 +21,31 @@
...
+        elif x_frame_opt == "Allow":
+            # if it is Allow, than there shouldn't be a
+            # X-Frame-Options header in the first place,
+            # so replace line with empty string
+            return ""
+        elif x_frame_opt == "Allow on same domain":
+            return base + "\"SAMEORIGIN\";"
+    else:
```
jtwaleson: I think

    return 'add _header X-Frame-Options %s' % setting

would be cleaner than using `base +`

start.py
```diff
@@ -21,6 +21,31 @@
...
+        # defaulting to allow all
+        return ""
+
+x_frame_opt = os.environ.get('X_FRAME_OPTIONS')
```
jtwaleson: one variable would be my preference, so:
    os.environ.get('X_FRAME_OPTIONS')
the value would contain the literal header value, and if None or 'Allow' it would not set any header.

jtwaleson:
looks good now :)

xiwenc:
codes looks good

PR #38, opened 2016-07-06
Update mxbuild-server for FastDeploy to build-10037

djvdorp:


lib/fastpush.py
```diff
@@ -0,0 +1,150 @@
...
+import subprocess
+
+
+ROOT_DIR = os.getcwd() + '/'
+MXBUILD_FOLDER = ROOT_DIR + 'mxbuild/'
+
+
+PROJECT_DIR = '.local/project'
+DEPLOYMENT_DIR = os.path.join(PROJECT_DIR, 'deployment')
+TMP_PROJECT_DIR = '.local/tmp_project'
```
riccardomc: These should have a more descriptive name.

lib/fastpush.py
```diff
@@ -0,0 +1,150 @@
...
+    def __init__(self, port, restart_callback, reload_callback, mx_version):
+        super(FastPushThread, self).__init__()
+        self.daemon = True
+        self.port = port
+        self.restart_callback = restart_callback
+        self.reload_callback = reload_callback
+        self.mx_version = mx_version
+
+    def run(self):
+        do_run(
```
riccardomc: Why do_run is outside of the class?

lib/fastpush.py
djvdorp: Might be good if we describe the workflow between MPK -> intermediate (tmp) directory -> deployment directory a bit more so we also know what, why and how in a few months from now?

PR #40, opened 2016-07-18
implement bluemix object storage

jtwaleson:
note that the "container" argument should probably be improved, it's a
user configurable namespace for objects, but there is no way for the
user to configure it here. Right now it we just use the name of the cf
service, but if that includes spaces or other characters this might
break. We could also just hard code it to "mendix" or something.

for this implementation a split between S3 based configuration and Swift
object store configuration needed to be introduced.

### DIFF
--- a/start.py
+++ b/start.py
@@ -194,13 +194,12 @@ def set_heap_size(javaopts):
     logger.debug('Java heap size set to %s' % max_memory)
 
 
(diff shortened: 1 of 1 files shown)


djvdorp:
CI failed because of a missing dependency, fixed that for you, LGTM!

PR #41, opened 2016-08-02
implements mono 4.4.1.0 supports

emirozer:
Please do not merge yet! :skull: 

We have to discuss which mx version we are going to support mono 4.4.1.0


lib/buildpackutil.py
```diff
@@ -275,25 +277,33 @@ def fix_mono_config_and_get_env(dot_local_location, mono_lib_dir):
...
-        '/usr/local/share/mono-3.10.0',
-        '/tmp/mono',
-    ], 'Mono not found')
+def _get_mono_path(dot_local_location, mendix_version):
+
+    if LooseVersion(mendix_version) < LooseVersion('6.8'):
+        return get_existing_directory_or_raise([
+            os.path.join(dot_local_location, 'mono'),
+            '/opt/mono-3.10.0',
+            '/tmp/mono',
```
xiwenc: I think it's nicer to have `/tmp/mono-3.10.0` as path because you know which version of mono is being used without calling the binary.

lib/buildpackutil.py
```diff
@@ -275,25 +277,33 @@ def fix_mono_config_and_get_env(dot_local_location, mono_lib_dir):
...
+        ], 'Mono not found')
+    else:
+        return get_existing_directory_or_raise([
+            os.path.join(dot_local_location, 'mono'),
+            '/opt/mono-4.4.1.0',
+            '/tmp/mono',
```
xiwenc: See above

lib/buildpackutil.py
```diff
@@ -304,16 +314,24 @@ def lazy_remove_file(filename):
...
-        return _get_mono_path(directory)
+        return _get_mono_path(directory, mendix_version)
     except NotFoundException:
-        download_and_unpack(
-            get_blobstore_url('/mx-buildpack/mono-3.10.0.tar.gz'),
-            directory,
-            cache_dir,
-        )
-    return _get_mono_path(directory)
+        if LooseVersion(mendix_version) < LooseVersion('6.8'):
```
xiwenc: define a function `def _get_mono_version(mendix_version)` and use that because there's code duplication

lib/buildpackutil.py
```diff
@@ -275,25 +277,33 @@ def fix_mono_config_and_get_env(dot_local_location, mono_lib_dir):
...
-    return get_existing_directory_or_raise([
-        os.path.join(dot_local_location, 'mono'),
-        '/usr/local/share/mono-3.10.0',
-        '/tmp/mono',
-    ], 'Mono not found')
+def _get_mono_path(dot_local_location, mendix_version):
+
+    if LooseVersion(mendix_version) < LooseVersion('6.8'):
+        return get_existing_directory_or_raise([
+            os.path.join(dot_local_location, 'mono'),
```
xiwenc: here also

emirozer:
changes finished, tests passing

lib/buildpackutil.py
```diff
@@ -122,9 +124,11 @@ def download_and_unpack(url, destination, cache_dir='/tmp/downloads'):
             ['dpkg-deb', '-x', cached_location, destination]
         )
     elif file_name.endswith('.tar.gz') or file_name.endswith('.tgz'):
-        subprocess.check_call(
-            ['tar', 'xf', cached_location, '-C', destination]
-        )
+        unpack_cmd = ['tar', 'xf', cached_location, '-C', destination]
+        if 'mono' in file_name:
```
xiwenc: it's better to make this a more strict condition: `filename.startswith("mono-")` because the current condition is too loose.

lib/buildpackutil.py
```diff
@@ -7,6 +7,7 @@
 import sys
 sys.path.insert(0, 'lib')
 import requests
+from distutils.version import LooseVersion
```
riccardomc: Is this necessary?

RaphaelAudet:
replaced

PR #43, opened 2016-08-12
S864950 dual mono version support

RaphaelAudet:
Hello Dear Maintainers

This pull request will change behavior while looking for the mono package. It will look for mono 3 or mono 4 based on the environment variable **FORCED_MONO4_VERSION**
```cf set-env application FORCED_MONO4_VERSION true```
By default, mono 3 will be used.

If mono is not found in expected location on the rootfs (/opt/mono-3.10.0 or /opt/mono-4.4.1.0), the buildpack will download the missing mono package from mendix cdn.

 The buildpack no longer updated the following libraries :
- libMonoPosixHelper.so
- libgdiplus.so
- libgdiplus.so.0
- libgdiplus.so.0.0.0
- libgif.so.4
- libgif.so.4.1
- libgif.so.4.1.6

These packaged version of mono on mendix cdn have been patched.

### CODE REVIEW

File: lib/buildpackutil.py [hunk 1]
```diff
@@ -278,25 +281,34 @@ def fix_mono_config_and_get_env(dot_local_location, mono_lib_dir):
...
+    logging.info('Targeting Mono Runtime: ' + target)
+    return target
+
+
+def _get_mono_path(dot_local_location, mendix_version):
+    mono_version = _get_mono_version(mendix_version)
     return get_existing_directory_or_raise([
-        os.path.join(dot_local_location, 'mono'),
-        '/opt/mono-3.10.0',
-        '/usr/local/share/mono-3.10.0',
```
djvdorp: AFAIK this line should stay here for backwards compatibility, since we specifically create a symlink while building the rootfs, see commit `0e8e5db423c9e57a4c641646fc51b6a47a701340` in Gitlab `cf-tools-bash`, file `mxlinuxfs2/mxrootfs.sh`.

lib/buildpackutil.py
```diff
@@ -265,38 +275,29 @@ def _checkout_from_git_rootfs(directory, mx_version):
...
+        target = 'mono-3.10.0'
+    logging.info('Selecting Mono Runtime: ' + target)
+    return target
+
+
+def _get_mono_path(directory, mono_version):
     return get_existing_directory_or_raise([
-        os.path.join(dot_local_location, 'mono'),
-        '/opt/mono-3.10.0',
-        '/usr/local/share/mono-3.10.0',
```
djvdorp: AFAIK this line should stay here for backwards compatibility with previous rootfs'es, since we specifically create a symlink while building the rootfs, see commit `0e8e5db423c9e57a4c641646fc51b6a47a701340` in Gitlab `cf-tools-bash`, file `mxlinuxfs2/mxrootfs.sh`.

lib/buildpackutil.py
```diff
@@ -265,38 +275,29 @@ def _checkout_from_git_rootfs(directory, mx_version):
...
-            mono_lib_dir, 'libMonoPosixHelper.so'
-        ),
-        os.path.join(_get_mono_path(dot_local_location), 'etc/mono/config'),
-    ])
     return env
 
 
-def _get_mono_path(dot_local_location):
+def _detect_mono_version():
+    if os.environ.get('FORCED_MONO4_VERSION'):
```
djvdorp: So, this effectively means we're never going to use Mono 4 based on a Mendix version, but only if we set the `FORCED_MONO4_VERSION` env var, correct?

lib/buildpackutil.py
```diff
@@ -265,38 +275,29 @@ def _checkout_from_git_rootfs(directory, mx_version):
...
+
+
+def _get_mono_path(directory, mono_version):
     return get_existing_directory_or_raise([
-        os.path.join(dot_local_location, 'mono'),
-        '/opt/mono-3.10.0',
-        '/usr/local/share/mono-3.10.0',
-        '/tmp/mono',
+        os.path.join(directory, mono_version),
+        '/opt/' + mono_version,
```
djvdorp: Why not also use `os.path.join` here, for this one and the next one?

lib/buildpackutil.py
```diff
@@ -308,20 +309,28 @@ def lazy_remove_file(filename):
             raise
 
 
-def ensure_and_get_mono(directory, cache_dir):
+def ensure_and_get_mono(mx_version, cache_dir):
+    logging.debug('ensuring mono for mendix {mx_version}'.format(
+        mx_version=mx_version
+    ))
+    mono_version = _detect_mono_version()
+    fallback_location = '/tmp/opt/' + mono_version
```
djvdorp: Why do we specifically want `/tmp/opt/` ?

tests/usecase/basetest.py
```diff
@@ -35,7 +35,7 @@ def setUpCF(self, package_name):
             "cf push -p %s -n %s %s --no-start -b https://github.com/mendix/cf-mendix-buildpack.git#%s" % (self.package_name, subdomain, self.app_name, self.branch_name),
             "cf create-service schnapps basic \"%s-schnapps\"" % self.app_name,
             "cf create-service PostgreSQL \"Basic PostgreSQL Plan\" \"%s-database\"" % self.app_name,
-            "cf create-service amazon-s3 basic \"%s-storage\"" % self.app_name,
+            "cf create-service amazon-s3 shared \"%s-storage\"" % self.app_name,
```
djvdorp: :+1:

lib/buildpackutil.py
```diff
@@ -265,38 +275,29 @@ def _checkout_from_git_rootfs(directory, mx_version):
     )
 
 
-def fix_mono_config_and_get_env(dot_local_location, mono_lib_dir):
+def _get_env_with_monolib(mono_lib_dir):
     env = dict(os.environ)
-    env['LD_LIBRARY_PATH'] = mono_lib_dir
-
-    if not os.path.isfile(os.path.join(mono_lib_dir, 'libgdiplus.so')):
+    env['LD_LIBRARY_PATH'] = mono_lib_dir + '/lib'
```
djvdorp: `os.path.join` might be preferred for this one and the next one, instead of using `+`

djvdorp:
Looks mostly good to merge, assuming this was tested in all possible scenarios as well? 
The comment regarding `/usr/local/share/mono-3.10.0` removal is the most urgent one.

lib/buildpackutil.py
xiwenc: for now yes, until we know for sure from which Mendix version on mono-4 is required. We use this flag to explicitly test mono-4

lib/buildpackutil.py
xiwenc: `os.path.join('', 'opt', mono_version)` notice the empty string as first argument

lib/buildpackutil.py
xiwenc: this is why the order of release is "important". If we release new cfrootfs then this change is fine. If not users will experience longer deployments but would still succeed.

lib/buildpackutil.py
RaphaelAudet: no, it's only to mimic the /opt tree

PR #49, opened 2016-08-29
Faster tests, add test for sampledata and fix sampledata issue

jtwaleson:


jtwaleson:
The travis-ci PR failed justly, because it always used cf push -b "https://github.com/mendix/cf-mendix-buildpack#master". The problem is that the merged code (between source branch and master) is only checked out locally in travis-ci. When you do "cf push -b" you have no way to target this merged code, you can only use the source branch or master.

We could push a temporary branch of the current merged source to somewhere so cf can download it, but that is ugly. Unchecked the "build on PR" option in travis ci.

Note: always rebase your branches before merging to master, otherwise the test might not be reflect the actual code that will be in master.

jtwaleson:
So actually, it is very good that this test failed, because the old buildpack (master) indeed had a problem with sampledata.

xiwenc:
looks good

PR #51, opened 2016-09-05
S864927 update m2ee

xiwenc:
- update to m2ee-0.5.11.3 (for java 8 graphs)
- refactored munin.py partially so that we can reuse the functions in buildpack


djvdorp:
LGTM!

PR #54, opened 2016-09-15
fastdeploy: if build contains errors, don't restart

jtwaleson:
also return a FAILED state so the WM knows there were errors


jtwaleson:
one capacity issue, the fastdeploy test succeeded

PR #55, opened 2016-09-16
690250 critical logs

riccardomc:
Introduce critical log message count in MENDIX-METRICS

* 	emit critical log message count within MENDIX-METRICS
* 	prevent emitter thread from exiting on exceptions

### DIFF
--- a/lib/metrics.py
+++ b/lib/metrics.py
@@ -14,17 +14,20 @@ class MetricsEmitterThread(threading.Thread):
     def run(self):
         logger.debug('Starting metrics emitter with interval %d' % self.interval)
         while True:
-            m2ee_stats, java_version = munin.get_stats_from_runtime(self.m2ee.client, self.m2ee.config)
-            m2ee_stats = munin.augment_and_fix_stats(
-                m2ee_stats,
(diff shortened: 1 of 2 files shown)


lib/metrics.py
```diff
@@ -14,15 +14,20 @@ def __init__(self, interval, m2ee):
...
-            }
-            logger.info('MENDIX-METRICS: ' + json.dumps(stats))
+            try:
+                m2ee_stats, java_version = munin.get_stats_from_runtime(self.m2ee.client, self.m2ee.config)
+                m2ee_stats = munin.augment_and_fix_stats(
+                    m2ee_stats,
+                    self.m2ee.runner.get_pid(),
+                    java_version)
+                critical_logs_count = len(self.m2ee.client.get_critical_log_messages())
+                m2ee_stats['critical_logs_count'] = critical_logs_count
```
jtwaleson: Maybe add it as another key in the stats object? So not under 'mendix_runtime'

riccardomc: Discussed with @xiwenc, we thought it was better this way.

jtwaleson: Ok, it doesn't really matter.

PR #60, opened 2016-11-07
Support for Mendix 7 cluster leader

arjenw:
Cluster support configuration has changed in Mendix 7. In case of a Mendix cluster, the instance with nr `0` is the cluster leader, while all other instances are the cluster slave.

Cluster leaders control database synchronization. Cluster slave instances wait for that to complete before accepting requests.

The Mendix 6 cluster options no longer apply to Mendix 7.

### DIFF
--- a/README.md
+++ b/README.md
@@ -131,7 +131,7 @@ If the setting contains a dot `.` you can use an underscore `_` in the environme
 
 ### Horizontal Scaling
 
-There are two ways for horizontal scaling in Mendix. In Mendix 5.15+ you can use sticky sessions or starting with Mendix 6 you can start Cluster Support (BETA). See below on how to activate both these settings.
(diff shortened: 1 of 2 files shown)


jtwaleson:
Looks good. Can we get rid of the old redis cluster mode support in the buildpack? I'm quite confident that it's not used by any of our customers on CF.

arjenw:
Getting rid of the old redis cluster mode support is up to you. Doing so means that no one can run any cluster on Mendix 6 using cloud foundry. If you think that is fine, feel free. But I rather decouple that from this PR as it doesn't impact each-other.

arjenw:
I've now also updated the documentation to include Mendix 7 clustering information. Please review (and merge :))

jtwaleson:
Cool, merged

PR #61, opened 2016-11-10
remove mx6 cluster option

jtwaleson:
This was a beta release and we move all the state to the client in Mendix 7. Let's remove this option to keep the buildpack simpler.


djvdorp:
CI failed because of disk staging issue, not related, merging since it looks good to merge to me!

PR #62, opened 2016-11-17
app user / admin port can have different passwords

jtwaleson:
if you want a different admin user password within the app, you can use
the APP_ADMIN_PASSWORD user. The /_mxbuild/ m2ee api will have the
ADMIN_PASSWORD value. This is not exposed to customers and thus not
documented, but we'll use it in the Mendix Cloud


djvdorp:
CI failed because of schnapps-service-broker issue, not related, merging since it looks good to merge to me!

PR #63, opened 2016-11-21
add instance_index to metrics emitter

emirozer:
Not defaulting to 0 here, it will mess with the data


emirozer:
The protobuf of the container metric includes instance index, ignore this pr

PR #65, opened 2016-12-13
failover for settings max memory and heap size

emirozer:
fix related to internal bug report: issue number 58

### DIFF
--- a/start.py
+++ b/start.py
@@ -203,14 +203,23 @@ def get_constants(metadata):
     return constants
 
 
-def set_heap_size(javaopts):
-    max_memory = os.environ.get('MEMORY_LIMIT', '512m').upper()
-    match = re.search('([0-9]+)([A-Z])', max_memory)
-    max_memory = '%d%s' % (int(match.group(1)) / 2, match.group(2))
-    heap_size = os.environ.get('HEAP_SIZE', max_memory)
+def set_heap_size(javaopts, vcap_max_mem):
+    max_memory = os.environ.get('MEMORY_LIMIT')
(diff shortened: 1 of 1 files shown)


xiwenc:
the commit message is not correct. It was already dynamic by default. The real change here is that `limits.memory` from `VCAP_APPLICATON` is used instead `MEMORY_LIMIT`.

emirozer:
I have changed the commit msg and the pr @xiwenc  / @djvdorp 
Let me know what you think..

start.py
```diff
@@ -203,14 +203,23 @@ def get_constants(metadata):
...
+
+    if max_memory:
+        max_memory.upper()
+        match = re.search('([0-9]+)([A-Z])', max_memory)
+        heap_size = '%d%s' % (int(match.group(1)) / 2, match.group(2))
+    else:
+        heap_size = str(int(vcap_max_mem) / 2) + 'M'
+
+    if env_heap_size:
+        heap_size = env_heap_size if int(env_heap_size[:-1]) < int(max_memory[:-1]) else heap_size
```
djvdorp: Inside this if, I think you might not have `max_memory`, what would happen then? I think it would break.
You might have to check that you have `max_memory` there also before trying to actually use it while it's `None`.

emirozer: yes it can be none! thx

emirozer: fixed

emirozer:
Also manually testing this, will report here

emirozer:
Manually tested heap determination with both HEAP_SIZE set and unset. Works as advertised :)

PR #66, opened 2016-12-16
fixes fallback condition where VCAP_SERVICES is not found

emirozer:
This was defaulting to None but None is not iterable so the following block of
get_database_url_from_vcap() (lib/buildpack.py:77) crashes.


djvdorp:
LGTM!

PR #68, opened 2016-12-27
M2EE_PASSWORD 

emirozer:
With this change, the newly deployed apps will simply won't have any user unless they do change admin pass.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,9 +31,7 @@ Also note that building the project in Cloud Foundry takes more time and require
 
 ### Configuring admin password
 
-The first push generates a new app, but deployment will fail because the buildpack requires an `ADMIN_PASSWORD` variable and a connected PostgreSQL or MySQL service. So go ahead and set this up after the first failed push.
-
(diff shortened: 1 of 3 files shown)


pommi:
Can we remove all of the `ADMIN_PASSWORD` stuff? Do we have to do `m2ee.client.create_admin_user` on startup? Maybe this can be removed as well? When a user needs to access it's App via MxAdmin, the admin password needs to be set anyway from the portal.

djvdorp:
I think `m2ee.client.create_admin_user` on startup might still be needed because of InstaDeploy? Not sure, just thinking out loud here, got notified via email about your comments haha.

emirozer:
I don't thin we can, users that want to run Mendix on other Cloud Foundry providers will have to set ADMIN_PASSWORD in the missing case of the M2EE_PASSWORD

@pommi @djvdorp

pommi:
Good point!

pommi:
The rest looks good.

PR #73, opened 2017-01-24
Added an example command to set the log level using cf client

RaphaelAudet:


pommi:
Sorry, reverted. Please use ` ``` ` around the command.

pommi:
I fixed it for you here: 2c971ebf92e0e758b961575d9b81558e94be4785

PR #76, opened 2017-02-14
Mark SERVER_ID as deprecated and use LICENSE_ID instead because this less confusing

xiwenc:
Internal processes refer `SERVER_ID` being the computed value of the `UUID` as input. To avoid confusion we will call this `UUID` value `LICENSE_ID` from now on.

### DIFF
--- a/tests/run.sh
+++ b/tests/run.sh
@@ -22,5 +22,9 @@ virtualenv -p python2 venv
 . venv/bin/activate
 pip install -r requirements.txt
 
+echo "Begin clean up of environment"
+cf apps | grep ops- | awk '{print $1}' | xargs -n 1 cf delete -r -f | true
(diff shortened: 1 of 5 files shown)


djvdorp:
@xiwenc can you ping me when this could be reviewed and merged?

xiwenc:
@jtwaleson do you approve?

tests/run.sh
```diff
@@ -22,5 +22,9 @@ virtualenv -p python2 venv
 . venv/bin/activate
 pip install -r requirements.txt
 
+echo "Begin clean up of environment"
```
jtwaleson: won't this remove the possibility to run multiple tests at the same time?

xiwenc: this removes possibility to run concurrent builds. But multiple tests are still possible.

PR #82, opened 2017-04-06
pick up s3_endpoint from s3 service broker to setup region specific e…

emirozer:
…ndpoint


jtwaleson:
We also support riak-cs and other S3 compatible file stores which don't support v4 signatures. So won't merge like this. You can add the enforcev4 ... java parameters from inside the service broker specific parser though.

PR #87, opened 2017-04-26
fix nginx logging filling up disk

jtwaleson:
This changes error_log to log to stderr natively, we didn't do anything
with the access log so just write that to /dev/null. Access logs are
handled by cloud foundry RTR logging already.


djvdorp:
LGTM!

PR #88, opened 2017-04-26
move from suffix to prefix for filedocuments

jtwaleson:
This is part of a long migration. We can now safely let the runtime
write filedocuments to prefix based objects in the s3 bucket.

The way we do this is hacky, but we need to support previous runtime
versions too. In the future the runtime can make a prefix option and
then we can have two code paths, one with the hacky way of setting the
prefix, one with the proper if runtime version is newer.

### DIFF
--- a/start.py
+++ b/start.py
@@ -20,7 +20,7 @@ from buildpackutil import i_am_primary_instance
 
 logger.setLevel(buildpackutil.get_buildpack_loglevel())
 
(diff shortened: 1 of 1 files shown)


djvdorp:
LGTM!

jtwaleson:
Failing tests are due to timeouts, nothing to do with these changes. Verified the functionality manually in test environment. Will merge now.

PR #93, opened 2017-05-23
emit request counters as absolute values

jtwaleson:
this will result in requests per second values for all request handlers


djvdorp:
LGTM!

PR #96, opened 2017-06-07
don't fill up logs on retrying ddl commands

jtwaleson:
Currently there were huge amounts of log messages if DDL commands failed. The
buildpack kept trying to execute the changes in a loop.


xiwenc:
looks good

PR #99, opened 2017-06-14
improves instadeploy responses being swallowed by response.raise_for_…

emirozer:
…status call

### DIFF
--- a/lib/instadeploy.py
+++ b/lib/instadeploy.py
@@ -34,6 +34,16 @@ for directory in (
     buildpackutil.mkdir_p(directory)
 
 
+class MxBuildFailure(Exception):
+    '''
+    Represents any 4xx 5xx issues retrieved from MxBuild HTTP Server
+    '''
+    def __init__(self, message, status_code, mxbuild_response):
+        super(MxBuildFailure, self).__init__(message)
+        self.status_code = status_code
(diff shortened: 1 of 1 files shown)


emirozer:
If the MxBuild Api return 4xx or 5xx we just discarded them silently with response.raise_for_status
this results in a bad experience of the instadeploy users. If their model is inconsistent or their java code doesnt compile they only see HTTPException  with 4xx client error or 5xx internal error. This is not helpful

lib/instadeploy.py
```diff
@@ -169,3 +187,13 @@ def build():
         ))
 
     return response.json()
+
+
+def extract_mxbuild_response(mxbuild_response):
+    r = {}
+    if mxbuild_response and 'problems' in mxbuild_response:
+        r['buildstatus'] = json.dumps(mxbuild_response['problems'])
+        if 'message' in mxbuild_response:
```
nobeh: I think this should have the same check and at the same level as `problems` and not nested inside:

```python
if mxbuild_response and 'message' in mxbuild_response:
```

emirozer: :+1:

emirozer: I also removed mxbuild_response None check as this is always going to be populated in the scope of this function

PR #100, opened 2017-06-14
Preserve the protocol by using 200 instead of error status

nobeh:
The user of the API receives 200 based on the previous implementation
but with more details in the response body. We change the `state` to
`FAILED` to remain consistent.


emirozer:
LGTM

PR #101, opened 2017-06-15
If starting app fails, record it at S3

nobeh:
Use the same approach as when MxBuild fails to build the package and
write the failure result into S3. This will ensure that Deployer always
gets a correct state from S3 whenever there's a failure either in
`compile` phase or `start` phase.

### DIFF
--- a/start.py
+++ b/start.py
@@ -965,9 +965,10 @@ if __name__ == '__main__':
         start_metrics(m2ee)
         start_nginx()
         loop_until_process_dies(m2ee)
-    except Exception:
+    except Exception as e:
         x = traceback.format_exc()
         logger.error('Starting app container failed: %s' % x)
         callback_url = os.environ.get('BUILD_STATUS_CALLBACK_URL')
         if callback_url:
             requests.put(callback_url, x)
+        raise e
--- a/start.py
+++ b/start.py
@@ -965,10 +965,10 @@ if __name__ == '__main__':
         start_metrics(m2ee)
         start_nginx()
         loop_until_process_dies(m2ee)
-    except Exception as e:
+    except Exception:
         x = traceback.format_exc()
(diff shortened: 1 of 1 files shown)


start.py
```diff
@@ -956,10 +957,17 @@ def sigterm_handler(_signo, _stack_frame):
...
+        set_up_instadeploy_if_deploy_password_is_set(m2ee)
+        start_metrics(m2ee)
+        start_nginx()
+        loop_until_process_dies(m2ee)
+    except Exception:
+        x = traceback.format_exc()
+        logger.error('Starting app container failed: %s' % x)
+        callback_url = os.environ.get('BUILD_STATUS_CALLBACK_URL')
+        if callback_url:
+            requests.put(callback_url, x)
```
jtwaleson: I think you want to raise the original exception here again at the end, via `raise`. Now only the error is logged but nothing crashes anymore, so Cloud Foundry doesn't know that the start failed.

Also, is there a good reason to try/except all these lines and not only around the `complete_start_procedure_safe_to_use_for_restart` call?

nobeh: Fair point on `raise`. I did not know that CF actually would receive it and
have an interpretation on it.

On second point, no. It was just a gut-feeling based on what I've learnt so
far. The intention is just to catch anything that can go wrong and record
the failure and proceed to crash.

Will update the PR.

nobeh: `raise` is done.

On the other one, so maybe I am missing something but we really do not know or cannot guarantee
that nothing will break. This is my intention that if anything breaks, record at S3 via Deployer such
that it should be still able to pick up `PackageState.FAILED` from CF and read from S3 about why.

If we wrap the exception handling around `complete_start_procedure_safe_to_use_for_restart`,
I am not sure we cover other failures.

PR #103, opened 2017-06-19
Preserve the previous `buildstatus` response field

nobeh:
We pass on the problems from MxBuild as `buildstatus` to the user of
instaDeploy.


emirozer:
:+1:

PR #106, opened 2017-06-21
Protect syncing files for failure and also note 'Busy' state

nobeh:


lib/instadeploy.py
```diff
@@ -184,16 +186,26 @@ def build():
     if response.status_code != requests.codes.ok:
         raise MxBuildFailure("MxBuild failure", response.status_code, response.json())
 
+    result = response.json()
+    if result['status'] == 'Success':
+        try:
+            sync_project_files()
+            logger.info("Syncing project files ...")
```
nobeh: From testability (which we do not have btw), the current change 
makes it easy to write a test case to be able to catch the issues from
`sync_project_files()`.

From readability, you are right. 

I leaned towards the first one. I can fix this in the next PR?

lib/instadeploy.py
```diff
@@ -184,16 +186,26 @@ def build():
...
+        except:
+            logger.warning("Syncing project files failed: %s", traceback.format_exc())
+    else:
+        logger.warning("Not syncing project files. MxBuild result: %s", result)
+
+    return result
+
+def sync_project_files():
```
nobeh: Motivation was small functions to make code more readable.

pommi: I do not agree here. Having many small functions will make things less readable because you constantly have to scroll through the code to see what is going on.

nobeh: That's why choosing readable function names are as important.

If we look from another side, we also have the cost and time for understanding the `for` loop
for somebody who wants to make a high-level understanding of the code. With using 
functions, I agree that the code becomes larger, but reading it would be like 
(a) do build (b) if success, do sync (c) otherwise raise error or issues

Without the function, I need to first read the code to see that `rsync` is used for
the purpose of syncing of project files after building.

This is my stand and perspective generally. I can change the code in the next PR
if you're still not convinced.

PR #110, opened 2017-07-03
exit improvements

jtwaleson:
the start.py, and with it nginx and the entire container did not die when the java process exited. This change improves that and adds a test case.


jtwaleson:
Ok, after refactoring some things in the test suite as well, this is good to go. @xiwenc @djvdorp  Please review and merge.

PR #111, opened 2017-07-12
improve logging about retrieving the Mendix Runtime version

pommi:


emirozer:
LGTM

emirozer:
@pommi the guys apparently want people to bump the version number in  start.py

https://github.com/mendix/cf-mendix-buildpack/blob/master/start.py#L26

djvdorp:
Looks good, but version number should be bumped and there is another change you want to do in `start.py` I think where we have similar logline(s) that could be more verbose, as you change did?

emirozer:
https://github.com/mendix/cf-mendix-buildpack/blob/master/start.py#L631

PR #112, opened 2017-07-22
DEP-62 Convert buildpack to Python3

jtwaleson:
https://mendix.atlassian.net/browse/DEP-62

python3 removes the SNI problems


lib/m2ee/client.py
```diff
@@ -36,7 +36,7 @@ def __init__(self, url, password):
         self._url = url
         self._headers = {
             'Content-Type': 'application/json',
-            'X-M2EE-Authentication': b64encode(password)
+            'X-M2EE-Authentication': b64encode(password.encode('utf-8')).decode('ascii'),
```
joshp123: Is there a reason we're decoding the b64 to ASCII instead of UTF-8?

Although, I would imagine that they are equivalent (because base64)

lib/m2ee/runner.py
```diff
@@ -25,23 +25,23 @@ def __init__(self, config, client):
...
             logger.warn("Cannot read pidfile: %s" % e)
             self._pid = None
 
     def _write_pidfile(self):
         if self._pid:
             pidfile = self._config.get_pidfile()
             try:
-                file(pidfile, 'w+').write("%s\n" % self._pid)
-            except IOError, e:
+                open(pidfile, 'w+').write("%s\n" % self._pid)
```
joshp123: We should add a context manager here, and put the write on a separate line.

lib/m2ee/runner.py
```diff
@@ -25,23 +25,23 @@ def __init__(self, config, client):
     def _read_pidfile(self):
         pidfile = self._config.get_pidfile()
         try:
-            pf = file(pidfile, 'r')
+            pf = open(pidfile, 'r')
```
joshp123: We should change this to use a context manager here.

lib/m2ee/client.py
djvdorp: +1 here also

PR #113, opened 2017-07-31
Update Java

pommi:


xiwenc:
Looks good to me. There's just a minor down side with dropping JRE: when MDA is pushed with the buildpack the droplet produced on non-mendix-cloud systems are bigger than JRE.

pommi:
@xiwenc Makes sense. Uncompressed JDK is ~160MB bigger then JRE. The commits I just pushed add back support for JRE.

djvdorp:
LGTM! Nice commit messages Pim!

jtwaleson:
I've rebased this onto master and added some changes. Let's continue in pull/118

PR #115, opened 2017-09-05
Get MySQL 2.0 uri from VCAP

jeroen-huizer-conclusion:
For MySQL 2.0 VCAP service name changed from p-mysql to p.msql.
See: http://docs.pivotal.io/p-mysql/2-0/index.html#AboutTile


xiwenc:
thanks for your contribution!

PR #116, opened 2017-09-07
Add logic to terminate child processes

nobeh:
On sys exit:

- Stop M2EE
- For every child process, try to terminate
- Allow each child to terminate gracefully for 1 second
- Wait maximum 3 seconds for all child processes
- Send SIGKILL in the end.

Signed-off-by: @emirozer

### DIFF
--- a/start.py
+++ b/start.py
@@ -9,21 +9,23 @@ import sys
 import base64
 import uuid
 import atexit
+import traceback
+
 sys.path.insert(0, 'lib')
 import requests
-from m2ee import M2EE, logger
 import buildpackutil
 import logging
 import instadeploy
 import metrics
+
+from m2ee import M2EE, logger
 from nginx import get_path_config, gen_htpasswd
 from buildpackutil import i_am_primary_instance
-import traceback
 
 logger.setLevel(buildpackutil.get_buildpack_loglevel())
 
 
-logger.info('Started Mendix Cloud Foundry Buildpack v1.5.1')
+logger.info('Started Mendix Cloud Foundry Buildpack v1.5.2')
 
 logging.getLogger('m2ee').propagate = False
 
@@ -122,13 +124,6 @@ def start_nginx():
     nginx_process = subprocess.Popen([
         'nginx/sbin/nginx', '-p', 'nginx', '-c', 'conf/nginx.conf'
     ])
-    atexit.register(stop_nginx)
-
-
-def stop_nginx():
-    if nginx_process:
-        logger.warning('Stopping nginx')
-        nginx_process.terminate()
 
 
 def get_vcap_data():
@@ -804,6 +799,24 @@ def start_app(m2ee):
     if abort:
         logger.warning('start failed, stopping')
         sys.exit(1)
+
+
+@atexit.register
+def terminate_process():
+    logger.info('stopping app...')
+    if not m2ee.stop():
+        if not m2ee.terminate():
+            m2ee.kill()
+    try:
+        this_process = os.getpgid(0)
+        logger.debug(
+            'Terminating process group with pgid={}'.format(this_process)
+        )
+        os.killpg(this_process, signal.SIGTERM)
+        time.sleep(3)
+        os.killpg(this_process, signal.SIGKILL)
+    except Exception:
+        logger.exception('Failed to terminate all child processes')
 
 
 def create_admin_user(m2ee):
--- a/tests/usecase/test_check_terminate_process_on_crashing_app.py
+++ b/tests/usecase/test_check_terminate_process_on_crashing_app.py
@@ -1,0 +1,19 @@
+import basetest
+
+
+class TestCaseTerminateChildProcessesCompleteOnCrashingApp(basetest.BaseTest):
+
+    def setUp(self):
+        self.setUpCF(
+            'Sample-StartError.mpk',
+            health_timeout=60,
+            env_vars={
+                'DEPLOY_PASSWORD': self.mx_password,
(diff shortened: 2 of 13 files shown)


xiwenc:
And a test case for this would be nice also

tests/usecase/test_check_terminate_process_on_crashing_app.py
```diff
@@ -0,0 +1,15 @@
...
+class TestCaseTerminateChildProcessesCompleteOnCrashingApp(basetest.BaseTest):
+
+    def setUp(self):
+        self.setUpCF('Sample-StartError.mpk', env_vars={
+            'DEPLOY_PASSWORD': self.mx_password,
+            'METRICS_INTERVAL': '10',
+        })
+        self.startApp()
+
+    def test_verify_termination_logs(self):
```
jtwaleson: I don't think this tests the actual error, which is that the state is still in starting mode, even though startup was aborted and start.py did exit(1).

My suggestion would be to check for the log line `start failed, stopping`, sleep 10 seconds and then see if instance state is not equal to `STARTING` in `cf curl /v2/apps/a0f891be-109f-4313-b542-7e5df6acf110/instances`

nobeh: Moved invocation of `startApp` from set-up to the test case method itself. 
The test case now also checks for the string when aborting is happen to stop the buildpack process.

start.py
```diff
@@ -805,6 +801,40 @@ def start_app(m2ee):
         logger.warning('start failed, stopping')
         sys.exit(1)
 
+@atexit.register
+def terminate_process():
```
jtwaleson: I don't think we should add the extra dependency (mostly because of shipping the .so file which will break if we switch python/linux versions).

Instead, we can kill the entire process group of the current process. I believe process groups were invented for purposes such as this. Example code:
```
>>> import os
>>> import signal
>>> os.killpg(os.getpgid(0), signal.SIGTERM)
```

Example process ids and groups for a container that we have:
```
vcap@8f18f639-cce0-4a2e-4745-18d18fa28d76:~$ ps  xao pid,ppid,pgid,sid,comm
  PID  PPID  PGID   SID COMMAND
    1     0     1     1 runc:[2:INIT]
    7     0     7     7 python
   11     0    11    11 diego-sshd
   60     7     7     7 sed
   62     0    61    61 java
  146     7     7     7 nginx
  147   146     7     7 nginx
 6845    11  6845  6845 bash
 6870  6845  6870  6845 ps
```

tests/usecase/test_check_terminate_process_on_crashing_app.py
```diff
@@ -0,0 +1,25 @@
...
+        try:
+            self.startApp(start_timeout=15)
+            self.fail("App should not have been started")
+        except subprocess.CalledProcessError:
+            pass
+        self.assert_string_in_recent_logs(self.app_name, 'start failed, stopping')
+        self.assert_string_in_recent_logs(self.app_name, 'Stopping M2EE ...')
+        self.assert_string_in_recent_logs(self.app_name, 'SIGTERM sent to all child processes.')
```
jtwaleson: I think we should only check the actual result of the problem (`health check never passed`) and leave out checks for the extra logging that we added in this PR.

start.py
```diff
@@ -805,6 +801,40 @@ def start_app(m2ee):
...
+    logger.warning('Stopping M2EE ...')
+    m2ee.stop()
+    logger.warning('Terminating all child processes ...')
+    try:
+        this_process = psutil.Process()
+        child_processes = this_process.children(recursive=True)
+        terminate_child_processes(child_processes)
+        logger.info('SIGTERM sent to all child processes.')
+        terminated_children, running_children = psutil.wait_procs(child_processes, timeout=3)
+        logger.info("Terminated {} child processes.".format(len(terminated_children)))
```
jtwaleson: Log lines on info will show up every single time a process is stopped. We should not bother the users with this. For debugging purposes while developing you can use the `BUILDPACK_XTRACE` environment variable.

start.py
nobeh: > `os.killpg(os.getpgid(0), signal.SIGTERM)`

This is a very strong and risky assumption that we always run a *containerized* environment such that we can kill parent process ID `0`.

nobeh: @jtwaleson All the comments were applied.

However, the test case for this PR consistently now takes the longest although we use a configured `start_timeout=15` with CF. Something to note down for later maybe.

nobeh: @jtwaleson @xiwenc 

Although the above changes are applied, there's something in the current code that needs to be observed. 

- Method `terminate_process` is registered as an exit handler for `sys`.
- Method `sigterm_handler` is registered as a SIGTERM handler [here](https://github.com/mendix/cf-mendix-buildpack/blob/master/start.py#L966).

Situation with `psutil`:

- :+1: A recursive list of child processes are obtained and ensured that all child processes are terminated and then `start.py` exists with `1`.
- :-1: A dependency is pulled in specifically for Linux that will not be portable if this buildpack is used in other platforms.

Situation with `os`:

- :+1: No external dependency.
- :+1: Send a standard `SIGTERM` to process group of `start.py`.
- :-1: Redundantly catch the same signal in the signal handler and do an exit `0`. This causes to lose the terminal exit code which is ` != 0`.

Using `os` creates a loop of logic because of using process group although it's simple. The API does not provide a way to exclude the parent process of the group. If we want to be precisely correct, the road map would be to implement a method which becomes similar to that of `psutil`.

PR #118, opened 2017-09-16
Update java

jtwaleson:


jtwaleson:
This is blocked until we update the rootfs in all the clusters.

pommi:
The new rootfs is deployed to all clusters. This can be merged.

PR #123, opened 2017-11-02
Adds postgresql_shared as database type

djvdorp:
As requested by @ernororive today:
This is to support a ‘postgres service broker’ which allows developers to divide a big postgres instance into smaller databases used by different apps.

Version should probably become v1.7.1. Bump and tag after merging.


pommi:
Maybe explain why in your commit message? Where does this "postgresql_shared" come from?

djvdorp:
@pommi thanks for the feedback, I have amended the commit to include this just now :+1:

PR #124, opened 2017-11-23
Flexible database service name

jtwaleson:


jtwaleson:
This ensures that anything that any service in VCAP that has a credentials -> uri string that contains postgres will be accepted as a database, so we don't have to hard code service names in the code anymore.

PR #129, opened 2018-01-04
Update start.py

MXClyde:
Print environment


MXClyde:
Unnecessary.

jtwaleson:
For future reference, it's better to use `BUILDPACK_XTRACE=true` for debugging things like this.

PR #130, opened 2018-01-08
fix mxbuild full deploy response formatting

jtwaleson:
For intadeploy we parse and do json.dumps on the mxbuild response too,
we should do the same for full deploy. Actually, we should probably
change the formatting froms string to actual json, but refactoring that
is a lot of effort for the consumers.

### DIFF
--- a/bin/compile
+++ b/bin/compile
@@ -240,10 +240,11 @@ def buildstatus_callback(error_file):
             "locations": []
         }
     ]}
-    if os.path.exists(error_file):
+    try:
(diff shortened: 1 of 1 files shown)


djvdorp:
LGTM!

xiwenc:
Not very clear whether formatted (pretty print) JSON is invalid JSON. Logically it shouldn't be an issue however according to https://www.json.org/ there may not be newlines outside of `string`. 

This change is semantically correct.

PR #132, opened 2018-01-10
fix reading past utf-8 bom

djvdorp:
See also #131, this is a follow-up.


xiwenc:
LGTM

djvdorp:
Relates to CLD-1591 and CLD-893.

PR #133, opened 2018-01-12
add option to apply rate-limiting on logging

pommi:
When environment variable LOG_RATELIMIT is set, logging from the Mendix
runtime gets piped through mendix-logfilter, that does 2 things:
* rate-limit loglines/second (defined in LOG_RATELIMIT)
* remove timestamp logged by Mendix (what the sed command does as well)


djvdorp:
Nice work Pim!

djvdorp:
Let's discuss this offline today, I was wondering if `LOG_RATELIMIT` shouldn't have a default value?

PR #139, opened 2018-02-19
Re: CLD-2188 Add metrics for crashes etc

hansthen:


xiwenc:
LGTM

start.py
```diff
@@ -1011,7 +1014,13 @@ def sigterm_handler(_signo, _stack_frame):
         m2ee.stop()
         sys.exit(0)
 
+    def sigusr_handler(_signo, _stack_frame):
+        metrics.emit(jvm = {'errors': float(_signo == signal.SIGUSR1),
+                            'ooms': float(_signo == signal.SIGUSR2)})
```
xiwenc: When this happens `loop_until_process_dies(...)` will terminate with another `emit(...)`. Confirmed with HansT

djvdorp:
See commit d649606 from @xiwenc. "1: kill SIGUSR2 7: not found"

Should probably have been this:
```
kill -SIGUSR1 MYPID
kill -SIGUSR2 MYPID
```

PR #143, opened 2018-02-21
set MALLOC_ARENA_MAX to a lower dynamic value to prevent OOM

jtwaleson:
The default behavior is 8x the number of detected CPUs . As Cloud
Foundry typically uses large host machines with smaller containers,
and the Java process is unaware of the difference in allocated CPUs,
the numbers are way off. This often leads to high native memory usage,
followed by a cgroup OOM killer event.

We go with Heroku's recommendation of lowering to a
setting of 2 for small instances. We also grown the setting linearly
with memory to be more in line with the default setting in Mendix
Cloud v3.

References:
- https://github.com/cloudfoundry/java-buildpack/issues/163
- https://devcenter.heroku.com/articles/testing-cedar-14-memory-use
- https://github.com/cloudfoundry/java-buildpack/issues/320

### DIFF
--- a/start.py
+++ b/start.py
@@ -241,7 +241,7 @@ def get_constants(metadata):
     return constants
 
 
-def set_jvm_memory(javaopts, vcap, java_version):
+def set_jvm_memory(m2ee_section, vcap, java_version):
     max_memory = os.environ.get('MEMORY_LIMIT')
 
     if max_memory:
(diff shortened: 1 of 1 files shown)


jtwaleson:
@hansthen in CF, the amount of CPU shares (not cores) scales linearly with the amount of memory. See also CLD-2455. I would also be fine with hard-coding to 2 for now.

knorrie:
Since we don't have any real life measurements of actual usage of these memory areas (as opposed to allocation of them) and/or malloc/free and lock contention rates in hosted applications, it makes sense to do anything to at least take the first step to lower the limits to sensible levels that correspond to cpu/memory sizing of the container limits, compared with running the same in a normal OS.

djvdorp:
What @knorrie said, +1 that.

jtwaleson:
discussed offline, we agreed on the current calculation

PR #145, opened 2018-02-28
fix access restrictions for /_mxbuild/ and /_mxadmin/

jtwaleson:
These locations are protected by basic auth only.

Not yet tested.


djvdorp:
LGTM!

djvdorp:
Peer-tested with Maarten, merging now.

PR #150, opened 2018-03-06
show a clear error when an invalid archive is pushed

jtwaleson:
Before this, we got sqlite database parameter problems because the
connection string was None.


djvdorp:
LGTM!

PR #151, opened 2018-03-09
Set DatabaseUseSsl to true if sslmode value is set.

xiwenc:
Official jdbc documentation:
sslmode = String
possible values include "disable", "require", "verify-ca" and
"verify-full", "allow" and "prefer" will throw an exception.
"require" will default to a non validating SSL factory and not
check the validity of the certificates. "verify-ca" and
"verify-full" use a validating SSL factory and will check that
the ca is correct and the host is correct.

Because we do not pass the original jdbc url we have to decompose the
url query params from `extra`. Also we need to map the possible sslmode
values into deciding to set `DatabaseUseSsl` to `True` or not.

source: https://jdbc.postgresql.org/documentation/head/connect.html

### DIFF
--- a/lib/buildpackutil.py
+++ b/lib/buildpackutil.py
@@ -4,6 +4,7 @@ import json
 import errno
 import subprocess
 import logging
+import urlparse
 import sys
 sys.path.insert(0, 'lib')
 import requests
@@ -53,6 +54,15 @@ def get_database_config(development_mode=False):
         'DatabaseHost': match.group('host'),
         'DatabaseName': match.group('dbname'),
     }
+
+    if 'extra' in match.groupdict() and match.group('extra'):
+        extra = match.group('extra').lstrip('?')
+        jdbc_params = urlparse.parse_qs(extra)
+        if 'sslmode' in jdbc_params:
+            sslmode = jdbc_params['sslmode']
+            if sslmode and sslmode[0] == 'require':
+                config.update({'DatabaseUseSsl': True})
+
     if development_mode:
         config.update({
             'ConnectionPoolingMaxIdle': 1,


lib/buildpackutil.py
```diff
@@ -53,6 +53,18 @@ def get_database_config(development_mode=False):
         'DatabaseHost': match.group('host'),
         'DatabaseName': match.group('dbname'),
     }
+
+    if 'extra' in match.groupdict():
+        extra = match.group('extra').lstrip('?')
+        import urlparse
```
joshp123: Shouldn't `import urlparse` be at the top of the file?

joshp123: Additionally, if this project is python3, we'll want to use `urllib.parse` instead of `urlparse`.

lib/buildpackutil.py
```diff
@@ -53,6 +53,18 @@ def get_database_config(development_mode=False):
         'DatabaseHost': match.group('host'),
         'DatabaseName': match.group('dbname'),
     }
+
+    if 'extra' in match.groupdict() and match.group('extra'):
+        extra = match.group('extra').lstrip('?')
+        import urlparse
```
joshp123: Shouldn't `import urlparse` be at the top of the file, along with the rest of the imports?

Additionally, if this project is python3, we'll want to use `urllib.parse` instead of `urlparse`.

lib/buildpackutil.py
```diff
@@ -53,6 +53,18 @@ def get_database_config(development_mode=False):
...
+        jdbc_params = urlparse.parse_qs(extra)
+        if 'sslmode' in jdbc_params:
+            use_ssl = ['verify-ca', 'verify-full',
+                       'allow', 'require', 'prefer']
+            sslmode = jdbc_params['sslmode']
```
joshp123: Since the output of `parse_qs` outputs a dict of [String, List], I don't think the if statement below will ever evaluate to true:
```
>>> qs = 'foo=bar&baz=qux'
>>> parse_qs(qs)
{'foo': ['bar'], 'baz': ['qux']}
>>>
```

`sslmode` will always be an array of strings, not a string, so when checking for membership in `use_ssl`, it will always fail.

Instead, I would suggest doing:
```
sslmode = jdbc_params['sslmode']
if sslmode and sslmode[0] in usessl:
   config.update({...})
```

lib/buildpackutil.py
```diff
@@ -53,6 +53,18 @@ def get_database_config(development_mode=False):
         'DatabaseHost': match.group('host'),
         'DatabaseName': match.group('dbname'),
     }
+
+    if 'extra' in match.groupdict():
+        extra = match.group('extra').lstrip('?')
+        import urlparse
+        jdbc_params = urlparse.parse_qs(extra)
+        if 'sslmode' in jdbc_params:
+            use_ssl = ['verify-ca', 'verify-full',
```
grootjans: When you set 'DatabaseUseSsl' to true, this implies 2 things. ssl=true and only a TLS connection without certificate validation. This behaves similar to sslmode=require. I think that sslmodes 'verify-ca' and 'verify-full' should not map to this, because they imply verification that is not done. 

Also, if I read the PostgreSQL docs correctluy, allow and prefer throw an exception.

xiwenc: So we should only support `require` which results in `DatabaseUseSsl` to `true`. Any other value ignore

xiwenc:
Customer confirmed fix works

PR #152, opened 2018-03-14
add auto-wiring for all mysql databases

jtwaleson:


jtwaleson:
Access Restrictions test is failing but is unrelated. It's likely to do with Travis CI enabling multiple IP addresses for their NAT gateway. @djvdorp can you review & merge?

PR #158, opened 2018-04-16
fix logs stopping after 1GiB

jtwaleson:
Currently, logs stop working after 1GiB because the runtime will rotate
the file and the fifo will break. Since we're logging to a fifo instead
of a regular file, we can just bump to a much higher value and not see
this problem again.


djvdorp:
LGTM!

PR #159, opened 2018-04-16
Introduce datadog (in alpha)

jtwaleson:
There are 4 main changes:
- add an agent which opens up statsd / opentracing endpoints, the
runtime can start to send metrics there
- add log subscriber on the runtime, the dd agent tails that file
- enable JMX on the mendix app, so JVM metrics are visible in datadog
- add a process agent, so all container processes CPU / max open files
stats etc are visible in datadog

All of this is currently in alpha and only activated after enabling
DD_API_KEY.

### DIFF
--- a/lib/datadog.py
+++ b/lib/datadog.py
@@ -62,7 +62,7 @@ def update_config(m2ee, app_name):
                 'run_path': '.local/datadog/run',
             },
         }
-        fh.write(yaml.dump(config))
+        fh.write(yaml.safe_dump(config))
     subprocess.check_call(('mkdir', '-p', '.local/datadog/conf.d/mendix.d'))
     subprocess.check_call(('mkdir', '-p', '.local/datadog/run'))
     with open('.local/datadog/conf.d/mendix.d/conf.yaml', 'w') as fh:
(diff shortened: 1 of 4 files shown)


lib/datadog.py
```diff
@@ -0,0 +1,113 @@
...
+        '-Dcom.sun.management.jmxremote.authenticate=false',
+        '-Dcom.sun.management.jmxremote.ssl=false',
+        '-Djava.rmi.server.hostname=127.0.0.1',
+    ])
+    m2ee.config._conf['logging'].append({
+        'type': 'file',
+        'name': 'FileSubscriberDataDog',
+        'autosubscribe': 'INFO',
+        'filename': os.path.join(os.getcwd(), 'log', 'datadog.log'),
+        'max_size': 1048576,
```
djvdorp: Shouldn't this also match #158 ?

lib/datadog.py
```diff
@@ -0,0 +1,113 @@
...
+import subprocess
+import buildpackutil
+
+
+def is_enabled():
+    return os.getenv('DD_API_KEY') is not None
+
+
+def _get_hostname():
+    domain = buildpackutil.get_vcap_data()['application_uris'][0].split('/')[0]
```
djvdorp: Isn't this a bit fragile?

lib/datadog.py
jtwaleson: No, here the runtime will actually handle file rotation because it's a regular file (not a fifo).

lib/datadog.py
jtwaleson: Yes, a bit, but good enough for alpha.

PR #160, opened 2018-04-24
add mx statsd java agent support (introduced in 7.14-rc4)

jtwaleson:
also refactor datadog configuration management a bit, and add the
DD_API_KEY env variables explicitly to both agents, to make
configuration more resilient

### DIFF
--- a/lib/datadog.py
+++ b/lib/datadog.py
@@ -5,8 +5,18 @@ import subprocess
 import buildpackutil
 
 
+def _get_api_key():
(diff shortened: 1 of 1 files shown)


jtwaleson:
The tests are very flaky due to Route53 issues. I triggered one more run, but otherwise it's good to merge as far as I'm concerned, all tests passed at least once.

djvdorp:
LGTM!

PR #167, opened 2018-05-17
DEP-142: Fix instadeploy

joshp123:
Also reverts the revert of #112 - the python 3 changes.

Use rebase and merge to ensure the history is clean.

https://mendix.atlassian.net/browse/DEP-142

PR #169, opened 2018-05-23
Better logging in tests

joshp123:
Always log the output of subprocess calls, this will help debugging non-deterministic tests.

~Based on #167, so do not merge until #167 is merged.~ That is now done.


jtwaleson:
I'm not sure, there is such a huge amount of output now that it makes it harder to debug things, at least for me.

joshp123:
We added `CF_TRACE` in 885003ba674c8057ad0f7d473ba7693af0f664a3. It is indeed too much, I have removed this.

PR #170, opened 2018-05-23
Add flake8 to buildpack.

joshp123:
This should improve our code quality both now and in the future.


joshp123:
Will replace with a rebased branch where we're using a 120chr limit.

PR #171, opened 2018-05-23
Add flake8

joshp123:
Depends on #169, but I think we can merge that too.

Flake8 is now in Travis CI, so code that does not pass will generate a unit test fail.

All files updated to pass flake8. We're using a 120chr limit because there's a fair few things that are ugly if you change them to be <80chr.

### DIFF
--- a/lib/metrics.py
+++ b/lib/metrics.py
@@ -1,17 +1,16 @@
+import datetime
+import json
 import os
 import sys
-import json
-import time
-from m2ee import logger, munin
-
 import threading
-import datetime
+import time
 
 BUILDPACK_DIR = os.path.dirname(os.path.dirname(os.path.realpath(__file__)))
-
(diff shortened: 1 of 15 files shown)


lib/metrics.py
```diff
@@ -149,7 +148,13 @@ def _get_database_mutations(self):
             db_config = buildpackutil.get_database_config()
             with conn.cursor() as cursor:
                 cursor.execute(
-                    "select xact_commit, xact_rollback, tup_inserted, tup_updated, tup_deleted from pg_stat_database where datname = '%s';" % (
+                    "SELECT xact_commit,"
+                    "       xact_rollback,"
+                    "       tup_inserted,"
+                    "       tup_updated,"
+                    "       tup_deleted"
+                    "FROM pg_stat_database"
```
jtwaleson: missing space before `"` in a couple of lines here?

joshp123: yes.

PR #178, opened 2018-06-07
DEP-2 client certificate validation

xiwenc:


joshp123:
Good to merge after CI passes.

PR #182, opened 2018-06-28
DEP-334: configure jetty options

xiwenc:
Manually tested:

```
cf set-env 675bc0be-de1a-4cc9-b067-78f0c8540a82 JETTY_CONFIG '{"runtime_max_threads": 100}'

```

and restarted the app:

screenshot 2018-06-28 at 13 33 24

### DIFF
--- a/start.py
+++ b/start.py
@@ -293,6 +293,19 @@ def set_jvm_memory(m2ee_section, vcap, java_version):
         )
 
 
+def set_jetty_config(m2ee):
+    jetty_config_json = os.environ.get('JETTY_CONFIG')
+    if not jetty_config_json:
+        return None
+    try:
+        jetty_config = json.loads(jetty_config_json)
+        jetty = m2ee.config._conf['m2ee']['jetty']
+        jetty.update(jetty_config)
+        logger.debug('Jetty configured: %s', json.dumps(jetty))
+    except Exception as e:
+        logger.warning('Failed to configure jetty', exc_info=True)
+
+
 def _get_s3_specific_config(vcap_services, m2ee):
     access_key = secret = bucket = encryption_keys = key_suffix = None
     endpoint = None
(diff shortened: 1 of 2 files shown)


start.py
```diff
@@ -293,6 +293,19 @@ def set_jvm_memory(m2ee_section, vcap, java_version):
...
 
+def set_jetty_config(m2ee):
+    jetty_config_json = os.environ.get('JETTY_CONFIG')
+    if not jetty_config_json:
+        return None
+    try:
+        jetty_config = json.loads(jetty_config_json)
+        jetty = m2ee.config._conf['m2ee']['jetty']
+        jetty.update(jetty_config)
+        logger.debug('Jetty configured: %s ' % json.dumps(jetty))
```
joshp123: Use the comma-based syntax for logging because otherwise it creates the string even when not required:
https://stackoverflow.com/a/34634301/2230226

tl;dr:
`logger.debug('foo is %s', foo)` vs `logger.debug('foo is %s' % foo)`
In #1 the string is lazily constructed, #2 is not.

tests/usecase/test_jetty_config.py
```diff
@@ -0,0 +1,16 @@
+import basetest
+import json
+
+
+class TestCaseJettyConfig(basetest.BaseTest):
```
joshp123: Can we also add tests for the unhappy paths?
- Invalid JSON in the `JETTY_CONFIG` environment variable
- (maybe) `JETTY_CONFIG` set but empty

start.py
```diff
@@ -293,6 +293,19 @@ def set_jvm_memory(m2ee_section, vcap, java_version):
...
+    except Exception as e:
+        logger.warning('Failed to configure jetty: ' + str(e))
```
joshp123: You can do this better in the following way:
```
logger.warning("Failed to configure jetty.", exc_info=True)
```
Because that includes not only the string representation of the exception, but also the traceback.

tests/usecase/test_jetty_config.py
xiwenc: added extra case with invalid json

pommi:
This was DEP-344 btw. not DEP-334

PR #183, opened 2018-07-13
Add Telegraf to the build pack. Download needs to go the CDN

hansthen:


lib/telegraf.py
```diff
@@ -0,0 +1,170 @@
+#
+# [EXPIRIMENTAL]
```
hansthen: EXPERIMENTAL

lib/telegraf.py
```diff
@@ -0,0 +1,170 @@
...
+def _get_appmetrics_target_url():
+    target = json.loads(_get_appmetrics_target())
+    if 'url' not in target:
+        logger.error('APPMETRICS_TARGET.url value is not defined in %s' % _get_appmetrics_target())
+    return target['url']
+
+
+def _get_appmetrics_target_username():
+    target = json.loads(_get_appmetrics_target())
+    return target['username'] if 'username' in target else None
```
hansthen: `return target.get('username')`?

lib/telegraf.py
```diff
@@ -0,0 +1,170 @@
...
+
+
+def _get_appmetrics_target_password():
+    target = json.loads(_get_appmetrics_target())
+    return target['password'] if 'password' in target else None
```
hansthen: `return target.get('password')`

lib/telegraf.py
```diff
@@ -0,0 +1,170 @@
...
+
+
+def _config_value_str(value):
+    if type(value) is str:
+        return '"%s"' % value
+    elif type(value) is int:
+        return value
```
hansthen: Is this correct?

cwesdorp: yes, for telegraf string values need to be quoted with double quotes.

lib/telegraf.py
```diff
@@ -0,0 +1,199 @@
+#
+# [EXPIRIMENTAL]
+#
+# Add Telegraf to an app container to collect StatsD events from the runtime.
+# Metrics will be forwarded to host defined in APPMETRICS_TARGET environment
+# variable which is a JSON (single or array) with the following values
+# - url: complate url of the endpoint. Mandatory.
```
xiwenc: `complate` -> `complete`

lib/telegraf.py
```diff
@@ -0,0 +1,199 @@
...
+    elif type(value) is int:
+        return value
+    elif type(value) is bool:
+        return str(value).lower()
+    elif type(value) is list:
+        return json.dumps(value)
+
+
+def _create_config_file(agent_config):
+    logger.info('writing config file')
```
xiwenc: Log as `debug` is preferred for this kind of cases to prevent excessively logging. Please revisit the other logging calls also.

cwesdorp: done

PR #187, opened 2018-08-05
Add check to datadog reporting DATABASE_DISKSTORAGE container env

cwesdorp:
Add information from the runtime to datadog so users can validate disk storage vs available storage given. This should enable checking disk (almost) full for the database.


xiwenc:
The commit only bumps up dd agent package version. How is this related to `DATABASE_DISKSTORAGE `?

cwesdorp:
The datadog package on CDN is updated to include a new script which reports back the value of DATABASE_DISKSTORAGE. It doesn't require additional config from the buildpack to activate this and when the env-variable is not present it will not send any information.

cwesdorp:
@joshp123  see commment in https://github.com/mendix/cf-mendix-buildpack/pull/188 on the version.

PR #188, opened 2018-08-10
Workaround telegraf basic auth issue by adding Authorization header

cwesdorp:
There is an issue with adding the authorization header to Telegraf. This is filed as https://github.com/influxdata/telegraf/issues/4544. Workaround is adding Authorization as custom header.

### DIFF
--- a/lib/telegraf.py
+++ b/lib/telegraf.py
@@ -11,10 +11,11 @@
 #     to custom end points
 #
 # Examples:
-# {"url: "https://customreceiver.mydomain.com", "username": "user", "password": secret", "kpionly": true}
-# [{"url: "https://customreceiver.mydomain.com", "username": "user", "password": secret", "kpionly": true}]
+# {"url": "https://customreceiver.mydomain.com", "username": "user", "password": "secret", "kpionly": true}
+# [{"url": "https://customreceiver.mydomain.com", "username": "user", "password": "secret", "kpionly": true}]
 #
 
+import base64
 import json
 import os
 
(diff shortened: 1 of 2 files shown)


lib/telegraf.py
```diff
@@ -103,8 +104,13 @@ def _write_http_output_config(http_config):
     username = http_config.get('username')
     password = http_config.get('password')
     if username is not None:
-        http_output['username'] = username
-        http_output['password'] = password
+        # Workaround for https://github.com/influxdata/telegraf/issues/4544
+        # http_output['username'] = username
+        # http_output['password'] = password
+        credentials = base64.b64encode(('%s:%s' % (username, password)).encode()).decode('ascii')
```
joshp123: Instead of using string interpolation, use .format.

So, in this case:

```
credentials = base64.b64encode("{}:{}".format(username, password).encode()).decode('ascii')
```

lib/telegraf.py
```diff
@@ -103,8 +104,13 @@ def _write_http_output_config(http_config):
...
+        http_output['[outputs.http.headers]'] = {
+            'Authorization': 'Basic ' + credentials
```
joshp123: Instead of using string concatenation, use `.format`.

start.py
```diff
@@ -27,7 +27,7 @@
 logger.setLevel(buildpackutil.get_buildpack_loglevel())
 
 
-logger.info('Started Mendix Cloud Foundry Buildpack v2.0.3')
+logger.info('Started Mendix Cloud Foundry Buildpack v2.1.2')
```
joshp123: Where are versions `v2.1.0` and `v2.1.1` ?

lib/telegraf.py
```diff
@@ -103,8 +104,13 @@ def _write_http_output_config(http_config):
     username = http_config.get('username')
     password = http_config.get('password')
     if username is not None:
```
joshp123: This wasn't changed here, but it's better to check for `if not username`.

If username happens to be the empty string, then the current statement (`is not None`) will evaluate to `True`, which is most likely not what you want.

start.py
cwesdorp: 2.1 was actually already released (without bumping the version number). Adding a new change for DD is 2.1.1 which makes this 2.1.2.

PR #189, opened 2018-08-16
Disable logging when posting to trends-storage-server fails.

joshp123:
Currently, customers end up with this message in their logs in cases
where there is no data loss. This causes unnecessary confusion.

We will replace this logging with grafana monitoring (DEP-87), and in
the long run, we should also allow the buildpack to send exceptions that
shouldn't be exposed to the customer via Sentry, so we can diagnose
them.


xiwenc:
looks good!

xiwenc:
alternatively change to `debug` and turn on debugging in ci test case

PR #190, opened 2018-08-17
WIP: DEP-677 Format all code with Black. Add black to CI.

joshp123:


joshp123:
Closing in favour of #198, which is based on a newer master.

PR #191, opened 2018-08-21
Fix encoding errors when using CERTIFICATE_AUTHORITIES

MXClyde:
Fix for regression introduced by upgrading to Python3


MXClyde:
Closing this. Can be fixed in Docker Build Pack.

PR #196, opened 2018-09-17
Re: DEP-463, DEP-511

hansthen:
Re: DEP-463 Add S3 storage metrics. This only works from a specific mendix
version, which is why we have the catch-all handler.

Re: DEP-511 Bandaid patcher to only print the first error message in case the logs
storage server is unavailable.


joshp123:
We already did DEP-511 here: https://github.com/mendix/cf-mendix-buildpack/pull/189

lib/metrics.py
```diff
@@ -30,14 +30,18 @@ class MetricsServerEmitter(MetricsEmitter):
...
     def emit(self, stats):
         try:
             response = requests.post(self.metrics_url, json=stats, timeout=10)
+            self.logfailure = True
         except Exception as e:
-            logger.debug(
-                "Failed to send metrics to trends server.", exc_info=True
-            )
+            if self.logfailure:
+                logger.warning(
```
joshp123: We already did DEP-511 here: https://github.com/mendix/cf-mendix-buildpack/pull/189, so the changes to the logging are not needed.

lib/metrics.py
```diff
@@ -285,6 +288,23 @@ def _get_number_of_files(self):
...
+                    "SELECT sum(size) from system$filedocument WHERE hascontents=true;"
+                )
+                rows = cursor.fetchall()
+                if len(rows) == 0:
+                    return 0
+                return int(rows[0][0])
+            except Exception as e:
+                # We ignore errors here, as the information is not available for
+                # older mendix versions
+                logger.debug("METRICS: Error retrieving file sizes", e)
```
joshp123: Since we are only logging this at `DEBUG` level, then it would be good to include the traceback, replace `e` with `exc_info=True`.

PR #197, opened 2018-09-17
Re: DEP-601 Add a heartbeat to the logs

hansthen:
This ensures no empty log files will exist for a day


joshp123:
This PR seems to contain lots of extraneous commits. Is the target branch a long way behind the source branch?

lib/metrics.py
```diff
@@ -57,6 +59,23 @@ def emit(self, stats):
...
 
 
+class LoggingHeartbeatEmitterThread(threading.Thread):
+    def __init__(self, interval):
+        super(LoggingHeartbeatEmitterThread, self).__init__()
+        self.interval = interval
+
+    def run(self):
+        logger.debug(
+            'Starting metrics emitter with interval %d' % self.interval
```
joshp123: For logging, the following syntax is best:

`logger.debug('Foo %s, bar)`, i.e. with the arguments to be merged in as arguments to logger.debug. (see: https://docs.python.org/3/library/logging.html#logging.Logger.debug)

This is better than plain string interpolation (or string concatenation, or `string.format`), as the arguments are only evaluated if the message is actually logged. (i.e. if your logging level is set higher, the string will not be constructed).

lib/metrics.py
```diff
@@ -57,6 +59,23 @@ def emit(self, stats):
             self.fallback_emitter.emit(stats)
 
 
+class LoggingHeartbeatEmitterThread(threading.Thread):
+    def __init__(self, interval):
+        super(LoggingHeartbeatEmitterThread, self).__init__()
```
joshp123: We use python3 now, so specifying the class and self in super is no longer needed, you can simply do:
`super().__init__()`

lib/metrics.py
```diff
@@ -57,6 +59,23 @@ def emit(self, stats):
...
+
+    def run(self):
+        logger.debug(
+            'Starting metrics emitter with interval %d' % self.interval
+        )
+        while True:
+            interesting_looking_string = ''.join(random.choice(
+                string.ascii_lowercase + string.digits
+            ) for _ in range(4))
+            logger.info('MENDIX-LOGGING-HEARTBEAT: ' + interesting_looking_string)
```
joshp123: As above re: logging syntax.

start.py
```diff
@@ -19,6 +19,7 @@
 import datadog  # noqa: E402
 import instadeploy  # noqa: E402
 import requests  # noqa: E402
+import metrics   # noqa: E402
```
joshp123: Metrics is part of the buildpack, so should be grouped with the 3rd group of imports (line 24 onwards), as per [PEP 8](https://www.python.org/dev/peps/pep-0008/?#imports)

joshp123:
Also if you format the code with black, then Travis CI will be happy! 😄

lib/metrics.py
hansthen: Done

start.py
hansthen: No longer relevant

lib/metrics.py
hansthen: Done

PR #198, opened 2018-09-19
DEP-677 Format all code with Black. Add Black to CI.

joshp123:
As per title.


Makefile
```diff
@@ -3,3 +3,4 @@ test: lint
 
 lint:
 	flake8 --exclude .git,__pycache__,lib/certifi,lib/idna,lib/psycopg2,lib/urllib3,lib/chardet,lib/httplib2,lib/requests,lib/yaml
+	black --check .
```
xiwenc: `--diff` includes why the check fails

PR #201, opened 2018-09-26
DEP-562: enable extra java locale providers for java8 or later

xiwenc:
See Jira issue for details

### DIFF
--- a/start.py
+++ b/start.py
@@ -228,6 +228,14 @@ def get_constants(metadata):
     return constants
 
 
+def set_jvm_locale(m2ee_section, java_version):
+    javaopts = m2ee_section["javaopts"]
+
+    # enable locale for java8 or later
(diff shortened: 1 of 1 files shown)


start.py
```diff
@@ -228,6 +228,14 @@ def get_constants(metadata):
     return constants
 
 
+def set_jvm_locale(m2ee_section, java_version):
+    javaopts = m2ee_section["javaopts"]
+
+    # enable locale for java8 or later
+    if not java_version.startswith("7"):
```
joshp123: The comment says "enable locale for java8 or later", but the code says "enable locale for java versions that are not 7".

Are we sure that the only Java versions are 7 and 8+?

xiwenc: yep, we do not support java older than 7 in buildpack

PR #204, opened 2018-10-04
CLDOPS-2524: Change heartbeat from Zen of Python to a generic counter.

joshp123:


cwesdorp:
Please update the version number as well so it will match the tag and release we will be creating.

cwesdorp:
Looks good, will merge this.

PR #205, opened 2018-10-04
DO: add DEBUG_CONTAINER mode

xiwenc:
This mode allows operators to start an application for troubleshooting the container. This saved a lot of time figuring out why the app was crashing with exit status code 143.

### DIFF
--- a/start.py
+++ b/start.py
@@ -15,6 +15,7 @@ import threading
 import time
 import traceback
 import uuid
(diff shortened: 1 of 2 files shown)


cwesdorp:
Don't merge this yet. If the environment variable is set the "cf start" command doesn't finish because the healthcheck can't determine if the container is ready (defaults to http). While the container is not healthy, you can't SSH into it.

PR #207, opened 2018-10-12
CLDOPS-2547 Supress errors on missing metrics data

joshp123:


djvdorp:
Nice fix, LGTM! Clean code.

xiwenc:
Note that this does not protect us from problems like `IndexError` because these get evaluated before `int_or_default(...)` gets called.

joshp123:
```

PR #209, opened 2018-10-25
Each test case use a dedicated CF_HOME

### LINKED ISSUE #156: Flaky test runs due to shared CF_HOME
Each test case should have its own dedicated `CF_HOME` to overcome possible concurrency issues in the cfcli:

- https://www.pivotaltracker.com/n/projects/892938/stories/128829195
- https://github.com/cloudfoundry/cli/issues/941

We should do this to improve the stability of test runs.

xiwenc:
Fixes #156

### DIFF
--- a/tests/usecase/basetest.py
+++ b/tests/usecase/basetest.py
@@ -4,6 +4,7 @@ import subprocess
 import unittest
 import uuid
 import requests
+import tempfile
 
 
 class BaseTest(unittest.TestCase):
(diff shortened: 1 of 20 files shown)


xiwenc:
Got 5 out of 6 good builds. The one that failed was due to "DNS resolution issue" which is something else that is flaky. This fixes the flakiness due to shared CF_HOME

PR #214, opened 2018-11-13
DEP-919 Change delete from storage to use newer setting.

joshp123:
See Jira ticket description.


xiwenc:
If I understand the issue correctly this change should only be applied if Mendix version > 7.19 is used. Else stick with old behavior.

joshp123:
Good catch,

PR #216, opened 2018-11-14
Fix download url for mxbuild

mweststrate:
https://download.mendix.com/runtime/mxbuild-7.11.0.tar.gz seems not to work
https://download.mendix.com/runtimes/mxbuild-7.11.0.tar.gz is working

(maybe this code path is never hit?)

### DIFF
--- a/lib/buildpackutil.py
+++ b/lib/buildpackutil.py
@@ -313,7 +313,7 @@ def ensure_mxbuild_in_directory(directory, mx_version, cache_dir):
             logging.debug(str(e))
             download_and_unpack(
                 get_blobstore_url(
-                    "/runtime/mxbuild-%s.tar.gz" % str(mx_version)
+                    "/runtimes/mxbuild-%s.tar.gz" % str(mx_version)
                 ),
                 directory,
                 cache_dir=cache_dir,


MXClyde:
Host used in the buildpack is https://cdn.mendix.com .
https://cdn.mendix.com/runtime/mxbuild-7.11.0.tar.gz works...

joshp123:
Clyde is correct, see here:
https://github.com/mendix/cf-mendix-buildpack/pull/216/files

I will close this MR; if you have a log of a failed deploy due to this issue, please re-open with said log. 🙂

mweststrate:
ah, thainks! The host I was using is download.mendix.com, where the same packages are available but under slightly different paths. Sorry for the confusion! 

I assume cdn.mendix.com is the preferred one for general usage?

joshp123:
No problem 🙂  `cdn.mendix.com` is indeed the preferred one. There are potentially changes to this in the pipeline, but SpecOps can tell you more about that than I can 🙂

c4rlier:
Hi @mweststrate sadly runtime dir was created under download.mendix.com and runtimes dir was created for cdn.mendix.com (i.e. someone did not check old locations when creating new ones on cdn).

Since the old path is used by on premise customers and in cloud v3 we can't just deprecate it without a long gruel-some process. When I get the time my goal is to point download.mendix.com to the CDN and keep the old paths as well. This means that you will be able to download the same file from runtime and runtimes dirs using download.mendix.com or cdn.mendix.com. If anyone wants to deprecate one of the URLs feel free to do it.

PR #217, opened 2018-11-28
CLDOPS-2680: increase proxy_read_timeout to 1h

djvdorp:
This is the same value that our frontfacing layer has set.
Putting it lower (eg. 15m) results in this when exceeding:

[error] 191#0: *82 upstream timed out (110: Connection timed out) while
reading response header from upstream

See also: https://stackoverflow.com/a/46205792/1007729


djvdorp:
@joshp123 done, see the next commit!

xiwenc:
great!

PR #218, opened 2018-11-30
DEP-758 Enable DataDog stasd listener besides Telegraf 

cwesdorp:
Enable DataDog stasd listener besides Telegraf to support Service Check Events from JMXFetch. Related to DEP-758.

### DIFF
--- a/lib/datadog.py
+++ b/lib/datadog.py
@@ -71,7 +71,10 @@ def update_config(m2ee, app_name):
         return
 
     tags = buildpackutil.get_tags()
-    statsd_port = 8125 if not buildpackutil.is_appmetrics_enabled() else 8126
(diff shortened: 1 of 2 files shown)


lib/datadog.py
```diff
@@ -71,6 +71,7 @@ def update_config(m2ee, app_name):
         return
 
     tags = buildpackutil.get_tags()
+    statsd_port = 8125 if not buildpackutil.is_appmetrics_enabled() else 8126
```
joshp123: I would rather we use an explicit if else here, instead of a conditional statement. (Also, the `if not` is less readable than simply using if`)

```suggestion
    if buildpackutil_is_appmetrics_enabled:
        statsd_port = 8126
    else:
        statsd_port = 8125
```

PR #222, opened 2018-12-17
Telegraf: MX Admin port and Azure insights

ronaldburgman:
My customer wants to monitor the Mendix runtime using Azure Application Insights.

These changes add an optional AAI configuration to telegraf if the APPMETRICS_AAI environment variable contains an instrumentation key. The changes also add http_inputs that monitor the Mendix admin port for statistics (e.g. heap memory) (depends on the latest telegraf version for the body parameter of the http_input plugin).

My customer prefers if this is implemented in the Mendix repository instead of in their own code. it also seems very useful for other customers that want to monitor the statistics provided by the admin port.

### DIFF
--- a/lib/telegraf.py
+++ b/lib/telegraf.py
@@ -212,12 +212,12 @@ def update_config(m2ee, app_name):
         _write_aai_output_config()
     
     # Collect statistics from Mendix admin port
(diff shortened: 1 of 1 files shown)


ronaldburgman:
Replaced this by #225 since I no longer seem to be able to edit this.

PR #223, opened 2018-12-21
DEP-1112: Include git commit hash in buildpack logging.

joshp123:
FYI, further to our discussion this morning, I have chosen not to include the branch name in the logging.

This is because the branch name is not a canonical reference, whereas a commit hash is. A single commit can exist in many branches, and thus the information is superfluous. The only information you need to identify the code at a given point in time is the commit hash.

Furthermore, customers should only ever be running the master branch, not any other branch. Nobody should ever force push to master (and indeed, this should be a protected branch in the repository), thus commit hashes in master should never change.

In the future, I indent for this code to be superseded by a more robust solution. This is described in DEP-1111.

### DIFF
--- a/bin/compile
+++ b/bin/compile
@@ -57,6 +57,32 @@ def check_environment_variable(variable, explanation):
         return True
 
 
+def get_current_git_commit():
+    try:
+        raw_commit = subprocess.check_output(
+            ["git", "rev-parse", "HEAD"], cwd=BUILDPACK_DIR
+        )
+        commit = raw_commit.decode("utf-8").strip()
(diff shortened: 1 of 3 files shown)


joshp123:
Thanks @pommi ! Do you think that this needs manual testing, or is peer review sufficient?

joshp123:
I added a unit test case for this. It is coupled to us using Travis CI, but I don't think that that should be a big concern here.

joshp123:
Rebased changes on the RELEASE_v2.3.0 branch.

PR #225, opened 2019-01-03
 Telegraf: MX Admin port and Azure insights

ronaldburgman:
My customer wants to monitor the Mendix runtime using Azure Application Insights.

These changes add an optional AAI configuration to telegraf if the APPMETRICS_AAI environment variable contains an instrumentation key. The changes also add http_inputs that monitor the Mendix admin port for statistics (e.g. heap memory) (depends on the latest telegraf version for the body parameter of the http_input plugin).

My customer prefers if this is implemented in the Mendix repository instead of in their own code. it also seems very useful for other customers that want to monitor the statistics provided by the admin port.

This has additional input metrics compared to #222. But I do not seem to be able to edit #222 anymore.

### DIFF
--- a/lib/telegraf.py
+++ b/lib/telegraf.py
@@ -29,8 +29,12 @@ def _get_appmetrics_target():
     return os.getenv("APPMETRICS_TARGET")
 
 
+def _get_appmetrics_aai():
+    return os.getenv("APPMETRICS_AAI")
+
+
 def is_enabled():
-    return _get_appmetrics_target() is not None
+    return _get_appmetrics_target() is not None or _get_appmetrics_aai() is not None
 
 
 def _is_installed():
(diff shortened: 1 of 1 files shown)


cwesdorp:
Hello Ronald,
 
We did check your merge request and decide to close it. Mendix provides the cf-mendix-buildpack and docker-mendix-buildpack as starting point to help you with deployment providing the Mendix supported services in it. Unfortunately we can't support all solutions in the standard build pack as there are lots of different configurations and combinations available which make it hard for us to guarantee we keep delivering a working build pack. We focus on a few options, you can use our work to extend this for your needs.
 
To keep your fork in sync you can checkout tools as Upriver or GitHubs Event API on repositories.

ronaldburgman:
Hi @cwesdorp 

In that case, can you at least add a way to skip the update_config call? Or add a check that it does not override an existing config file?

I do not mind adding a docker layer with my own configuration file (I understand you cannot account for every use case). But right it is not possible since this file is always overwritten when the container starts. Adding a docker layer is preferred because then I do not need to keep any code in sync.

PR #229, opened 2019-02-28
DEP-1391 Add AdoptOpenJDK download option, make the version comparision strate…

svanderburg:
…gy more robust

### DIFF
--- a/lib/buildpackutil.py
+++ b/lib/buildpackutil.py
@@ -10,6 +10,8 @@ sys.path.insert(0, "lib")
 
 import requests  # noqa: E402
 
+from m2ee.version import MXVersion  # noqa: E402
+
 
 def get_vcap_services_data():
     if os.environ.get("VCAP_SERVICES"):
@@ -96,7 +98,7 @@ def download_and_unpack(url, destination, cache_dir="/tmp/downloads"):
     )
     if file_name.endswith(".tar.gz") or file_name.endswith(".tgz"):
         unpack_cmd = ["tar", "xf", cached_location, "-C", destination]
-        if file_name.startswith(("mono-", "jdk-", "jre-")):
+        if file_name.startswith(("mono-", "jdk-", "jre-", "AdoptOpenJDK-")):
             unpack_cmd.extend(("--strip", "1"))
         subprocess.check_call(unpack_cmd)
     else:
@@ -158,10 +160,15 @@ class NotFoundException(Exception):
 
 
 def get_java_version(mx_version):
-    versions = {"7": "7u80", "8u51": "8u51", "8": "8"}
-    if mx_version >= 6.6:
-        default = "8"
-    elif mx_version >= 5.18:
+    versions = {"7": "7u80", "8u51": "8u51", "8": "8", "8u202": "8u202"}
+
+    if mx_version >= MXVersion("8"):
+        default = "8u202"
+    elif mx_version >= MXVersion("7.23.1"):
+        default = "8u202"
+    elif mx_version >= MXVersion("6.6"):
+        default = "8"  # This is actually version 8u144
+    elif mx_version >= MXVersion("5.18"):
         default = "8u51"
     else:
         default = "7"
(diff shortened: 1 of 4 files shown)


lib/buildpackutil.py
```diff
@@ -7,6 +7,7 @@
 import sys
 from distutils.util import strtobool
 from urllib.parse import parse_qs
+from m2ee.version import MXVersion
```
joshp123: Please follow pep8 and put this in a separate group of imports for first-party modules (insert 1 line above this)

lib/buildpackutil.py
```diff
@@ -433,23 +445,61 @@ def ensure_and_get_mono(mx_version, cache_dir):
...
+def _compose_jdk_url_path(java_version):
+    oracle_jdks = ["7", "8u51", "8"]
+    adoptopenjdk_jdks = ["8u202", "11"]
+
+    if java_version in oracle_jdks:
+        jdk_type = "jdk"
+    elif java_version in adoptopenjdk_jdks:
+        jdk_type = "AdoptOpenJDK"
+    else:
+        raise Exception("Unknown java version identifier: %s" % java_version)
```
joshp123: Please use `str.format` instead of string interpolation

lib/buildpackutil.py
```diff
@@ -433,23 +445,61 @@ def ensure_and_get_mono(mx_version, cache_dir):
...
+
+    return "/mx-buildpack/%s-%s-linux-x64.tar.gz" % (jdk_type, java_version)
```
joshp123: Please use `str.format` instead of string interpolation

lib/buildpackutil.py
```diff
@@ -433,23 +445,61 @@ def ensure_and_get_mono(mx_version, cache_dir):
...
+def _compose_jvm_target_dir(java_version):
+    oracle_jdks = ["7", "8u51", "8"]
+    adoptopenjdk_jdks = ["8u202", "11"]
+
+    if java_version in oracle_jdks:
+        jdk_type = "oracle"
+    elif java_version in adoptopenjdk_jdks:
+        jdk_type = "AdoptOpenJDK"
+    else:
+        raise Exception("Unknown java version identifier: %s" % java_version)
```
joshp123: Please use `str.format` instead of string interpolation

lib/buildpackutil.py
```diff
@@ -433,23 +445,61 @@ def ensure_and_get_mono(mx_version, cache_dir):
...
+
+    return "usr/lib/jvm/jdk-%s-%s-x64" % (java_version, jdk_type)
```
joshp123: Please use `str.format` instead of string interpolation

lib/buildpackutil.py
```diff
@@ -433,23 +445,61 @@ def ensure_and_get_mono(mx_version, cache_dir):
     return mono_location
 
 
+def _compose_jdk_url_path(java_version):
```
joshp123: The code in `_compose_jdk_url_path`, `_compose_jvm_target_dir` and `_compose_jre_url_path` are almost identical, but just return slightly different strings. Can we abstract the validation part into a separate function and minimize the amount of copy/pasted code?

lib/buildpackutil.py
joshp123: bump

start.py
```diff
@@ -66,7 +66,7 @@ def get_current_buildpack_commit():
 
 
 logger.info(
-    "Started Mendix Cloud Foundry Buildpack v2.2.13 [commit:%s]",
+    "Started Mendix Cloud Foundry Buildpack v2.3.0 [commit:%s]",
```
cwesdorp: Needs update, please check master branch for latest version

cwesdorp:
This looks good

tests/usecase/test_jdk_versions.py
```diff
@@ -0,0 +1,113 @@
...
+                        "java",
+                    )
+                    + " -version",
+                )
+            )
+            localCheckFail = False
+        except Exception:
+            localCheckFail = True
+
+        assert not (rootCheckFail and localCheckFail)
```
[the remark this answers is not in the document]
joshp123: I asked Matyas about this, because he's a bit more informed than I am:

> Generally I would go for the latter and make use of the dedicated assertions (eg. `assertListEqual`, `assertIsNone`, etc. - https://docs.python.org/3/library/unittest.html#assert-methods) that are available when using `unittest.TestCase` based test classes. Because _tests are specifications_, so readability and communicating intent is better (that’s why if there are more complex asserts, creating well-named custom assertion methods is good practice).
> 
> However if your test suite is written with the `pytest` style (eg. no test classes just methods), then you don’t have much choice but user plain `assert` statements. Although `pytest` is doing really good when dealing with these and reporting failures.
> 
> My personal preference is to stick to `unittest` features and keep `pytest` as a test-runner tool as much as possible. The reason is that even though `pytest` also provides everything you’d need, because of the plain `assert`s, readability suffers. The fixture functionality is extremely powerful, but that can also set back readability as sometimes you have to hunt around for fixture definitions as they are automagically injected, runtime, no imports or anything. (I actually had a hard time this morning when looking at an example cookiecutter project to find where a test fixture is coming from only to find out that it’s a `pytest` plugin with a single explicit occurrence in the requirements file, nowhere else).
> Furthermore you lock-in with `pytest` while in the future you might want to switch test-runners (parallelization, better reporting, etc.). (edited)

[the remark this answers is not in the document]
joshp123: So, tl;dr, please continue using the current unit test style, but transition to the `self.assert` methods.

lib/buildpackutil.py
```diff
@@ -328,8 +328,8 @@ def ensure_and_get_mono(mx_version, cache_dir):
 
 
 def _determine_jdk(mx_version, package="jdk"):
-    oracle_jdks = ["7", "8u51", "8"]
-    adoptopenjdk_jdks = ["8u202"]
+    oracle_jdks = ("7", "8u51", "8")
+    adoptopenjdk_jdks = "8u202"
```
joshp123: You changed this from a `list` with one element into simply a string, so your check on line `337`, `elif java version in adoptopenjdk_jdks` will not work as intended.

PR #230, opened 2019-03-03
start.py: rm "The Zen of Python" from the logging heartbeat message

pommi:
Let's just stop this madness. It causes unnecessary support tickets.
We have better things to do.

### DIFF
--- a/start.py
+++ b/start.py
@@ -1,9 +1,7 @@
 #!/usr/bin/env python3
 import atexit
 import base64
-import codecs
 import datetime
-import itertools
 import json
 import logging
 import os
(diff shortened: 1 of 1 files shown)


cwesdorp:
Indeed, let's remove it. 
Please update the version number in start.py as well and it is okay to merge.

start.py
```diff
@@ -1195,8 +1169,8 @@ def run(self):
         logger.debug(
             "Starting metrics emitter with interval %d", self.interval
         )
-        for line in itertools.cycle(HEARTBEAT_STRING_LIST):
-            logger.info("MENDIX-LOGGING-HEARTBEAT: %s", line)
+        while True:
+            logger.info("MENDIX-LOGGING-HEARTBEAT: .")
```
joshp123: I think the rationale here was to be able to distinguish lines, hence the choice of the Zen of Python (a stupid choice). I would suggest adding an incrementing counter here.

joshp123: Something like this should do it: https://github.com/mendix/cf-mendix-buildpack/pull/204

cwesdorp: Wasn't the idea behind the counter to have a recognisable value in the logs to make a calculation/estimation on the number of lines missed in the final files. I think we should define what we are actually going to do with this heartbeat information or should keep it simple as just a heartbeat.

joshp123: The original rationale, as I recall it, was to have a well-known message that repeated, so that the output of `cat logs | grep MENDIX-LOGGING-HEATBEAT` would illustrate if a message had been emitted every N hours, and then you could use this (vague) information to diagnose issues in the container (and e.g. figure out if the container had just been started or if it had been running for a long time, based purely on the logs, not `cf events`).

Anyway, the string-based approach, rotating through a list of N strings, is clearly sub-optimal. A counter would at least add some clarity to situations when e.g. hearbeats 3 and 5 are present, but 4 is missing. You could also figure this out based on extrapolating from timestamps, but its harder. Explicit is better than implicit, etc. (Ironically from the Zen of Python)

cwesdorp: Or remove it all. If the app is not logging it doesn't log. Recent logs can be viewed in live-logging (later) or using "cf logs --recent". It's not that we need the line.

joshp123: I disagree, the value of this is that it conclusively proves that, for a given (coarse) time period, the app was indeed running and logging.

The output `cf logs --recent` doesn't give any guarantees regarding retention. Live logging is only for now onwards. The use case here is proving that the application is indeed running, and showing whether or not the logging infra is running. I have found the presence of the logging heartbeat useful on numerous occasions, since you cannot get this information elsewhere.

pommi: What about logging a timestamp?

pommi:
Duplicate of #204

PR #232, opened 2019-03-05
UP-760 Allow the buildpack to use user provided JAVA_OPTS

diogomatsubara:
This change allows the user to set the JAVA_OPTS env var and the buildpack will unpack the value and add to the existing java options. This is was required to help with debugging with UP-760

### DIFF
--- a/start.py
+++ b/start.py
@@ -314,6 +314,13 @@ def set_jvm_locale(m2ee_section, java_version):
         javaopts.append("-Djava.locale.providers=JRE,SPI,CLDR")
 
 
+def set_user_provided_java_options(m2ee_section):
+    javaopts = m2ee_section["javaopts"]
+    options = os.environ.get("JAVA_OPTS", None)
+    if options:
+        javaopts.append(options.split())
+
+
 def set_jvm_memory(m2ee_section, vcap, java_version):
     max_memory = os.environ.get("MEMORY_LIMIT")
 
@@ -833,6 +840,7 @@ def set_up_m2ee_client(vcap_data):
     )
     set_jvm_memory(m2ee.config._conf["m2ee"], vcap_data, java_version)
     set_jvm_locale(m2ee.config._conf["m2ee"], java_version)
+    set_user_provided_java_options(m2ee.config._conf["m2ee"])
     set_jetty_config(m2ee)
     activate_new_relic(m2ee, vcap_data["application_name"])
     activate_appdynamics(m2ee, vcap_data["application_name"])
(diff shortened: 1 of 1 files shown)


start.py
```diff
@@ -300,6 +300,13 @@ def set_jvm_locale(m2ee_section, java_version):
         javaopts.append("-Djava.locale.providers=JRE,SPI,CLDR")
 
 
+def set_user_provided_java_options(m2ee_section):
+    javaopts = m2ee_section["javaopts"]
+    options = os.environ.get("JAVA_OPTS", None)
```
joshp123: What is the desired input format here?

It looks to me like a space-separated string, but I think that we can be a bit safer/more explicit here.

diogomatsubara: I'm not sure I understand your suggestion. Do you suggest that we accept JSON as the value of JAVA_OPTS? So something like:

cf set-env my_app JAVA_OPTS '["-XX:+UseG1GC", "-XX:+HeapDumpOnOutOfMemoryError"]'

So that I can then use this in json.loads() to validate it?

joshp123: Yes, exactly. Sorry for wording this poorly, we were in a hurry at the time.

(Note that I suggest using a list-type, not a dict-type here, which is uncommon but still valid JSON.)

diogomatsubara: I made the changes as you suggested but I think this makes the option harder to use and the validation that is being done give a false sense of security (in terms of values being actually validated or not).

Compare:
`cf set-env JAVA_OPTS "-XX:+UseG1GC -XX:+HeapDumpOnOutOfMemoryError"`
vs
`cf set-env JAVA_OPTS '["-XX:+UseG1GC", "-XX:+HeapDumpOnOutOfMemoryError"]'`

The additional bracket and quotes makes it harder to get it right, IMHO and in case the user use a non-existent java option or make a typo in one of the options, the java process will still fail to start.

joshp123: I see your point, and I'm inclined to agree if we're setting these options via the CF CLI.

However, I would imagine that java opts will (fairly soon) end up being set via Cloud Portal/Deployer. After all, manually setting java options on a per-container basis doesn't scale. Thus, I think that it's better to go with the more machine-friendly option (JSON) over the whitespace separated string.

Do you agree?

diogomatsubara: Not really. The use case I have in mind is to set this using the CF CLI to help debugging specific issues. I don't think customers can set those values by themselves and IIUC, when they set custom env vars through CP, CP only allows whitelisted env vars to be passed on to the buildpack. I'm not sure if there's a customer need/request to be able to set those themselves and we probably need to be careful allowing people to set custom options to JVM as that might make their app unstable and increase support tickets for things that we don't even have control over.

diogomatsubara: In any case, the code is done as you suggest. If you think this is good enough, please approve, otherwise I'll revert it back to the whitespace. I don't really mind the json that much, trying to make my life a little bit easier, but I will survive if I have to pass the extra option as JSON

joshp123:
Approved once you can get the CI passing.

PR #233, opened 2019-03-11
Upgrade buildpack libraries to resolve security findings

cwesdorp:
Upgrade buildpack lib/* libraries to resolve security findings. Also update test/requirements.txt


Makefile
```diff
@@ -2,5 +2,5 @@ test: lint
 	cd tests && bash run.sh
 
 lint:
-	flake8 --exclude .git,__pycache__,lib/certifi,lib/idna,lib/psycopg2,lib/urllib3,lib/chardet,lib/httplib2,lib/requests,lib/yaml
+	flake8 --exclude .git,__pycache__,lib/certifi,lib/idna,lib/psycopg2,lib/urllib3,lib/chardet,lib/httplib2,lib/requests,lib/yaml,lib/m2ee
```
joshp123: Why are we excluding m2ee from being checked by flake8? This is our code, so should conform to our coding standards.

lib/buildpackutil.py
```diff
@@ -479,7 +479,7 @@ def bypass_loggregator_logging():
     # Necessary since we store these in cloud portal as strings.
     try:
         bypass_loggregator = strtobool(env_var)
-    except ValueError as e:
+    except ValueError:
```
joshp123: The commit messages says that we are just updating libraries. But this is non-library code. If we are fixing linting issues, can we do this in a separate commit, please?

lib/metrics.py
```diff
@@ -19,7 +19,7 @@
 def int_or_default(value, default=0):
     try:
         return int(value)
-    except Exception as e:
+    except Exception:
```
joshp123: The commit messages says that we are just updating libraries. But this is non-library code. If we are fixing linting issues, can we do this in a separate commit, please? (This goes for all the changes in this file)

lib/nginx.py
```diff
@@ -32,7 +32,7 @@ def get_path_config():
     Default for satisfy is any
     """
 
-    location_template = """
+    location_template = r"""
```
joshp123: Changing a normal multi-line string to a raw string is a functional change. Why is this done along with updating the dependencies? And, what is the rationale for this change?

lib/nginx.py
```diff
@@ -45,7 +45,7 @@ def get_path_config():
     %s
 }
 """
-    root_template = """
+    root_template = r"""
```
joshp123: Changing a normal multi-line string to a raw string is a functional change. Why is this done along with updating the dependencies? And, what is the rationale for this change?

start.py
```diff
@@ -280,7 +280,7 @@ def get_constants(metadata):
     )
     try:
         constants_from_json = json.loads(constants_json)
-    except Exception as e:
+    except Exception:
```
joshp123: As above, linting related changes in an unrelated commit/PR. Please move to a separate commit/PR.

start.py
```diff
@@ -883,7 +883,7 @@ def set_up_logging_file():
             [
                 "sed",
                 "--unbuffered",
-                "s|^[0-9\-]\+\s[0-9:\.]\+\s||",
+                r"s|^[0-9\-]\+\s[0-9:\.]\+\s||",
```
joshp123: As above, normal string literal to a raw string is a functional change. What is the rationale for this, and why is it grouped with the library upgrades commit?

tests/requirements.txt
```diff
@@ -1,11 +1,11 @@
-flake8==3.5.0
```
joshp123: It's important to know that if you are changing the version of the linter, the results of linting will often change accordingly. It's important to decide if that's what you actually want.

Makefile
cwesdorp: Flake8 upgrade gives some new errors on m2ee. I excluded m2ee as it be the copy of 'pip install m2ee' but it's not (yet). I made the choice not to modify imported libraries.

joshp123: I don't think that's the right choice here. The version of m2ee used in the buildpack is divergent from the upstream version of m2ee, so reconciling the two is not something that will happen any time soon. In the meantime, any m2ee changes should conform to our coding standards.

I would prefer that we either:
a) fix m2ee to comply with this version of flake8
b) skip updating flake8 (it's a linter, so unrelated to any security concerns)

cwesdorp:
Removed the flake8 upgrade for now. We can do it separate indeed. For m2ee-tools we opened a ticket move to the vanilla release.

Makefile
cwesdorp: no longer in mr. opened a ticket to move the vanilla m2ee release.

lib/buildpackutil.py
cwesdorp: no longer in mr.

lib/metrics.py
cwesdorp: no longer in mr.

lib/nginx.py
cwesdorp: no longer in mr.

lib/nginx.py
cwesdorp: no longer in mr.

start.py
cwesdorp: no longer in mr.

tests/requirements.txt
cwesdorp: no longer in mr.

start.py
cwesdorp: no longer in mr.

lib/metrics.py
joshp123: You resolved this conversation but it does appear to actually still be in the MR.

lib/nginx.py
joshp123: You resolved this conversation but it does appear to actually still be in the MR.

PR #238, opened 2019-03-26
Add support for SAP HANA

cwesdorp:
Add support for SAP HANA when bound via cf service.
Refactored database code into separate file allowing for more fine grained control per database type. First implementation is for Hana.

### DIFF
--- a/docs/supported_database_vcaps.md
+++ b/docs/supported_database_vcaps.md
@@ -1,12 +1,12 @@
 # Overview of supported VCAP schemas
 
-Using Cloud Foundry database services are detected from VCAP bindings and translated into Mendix M2EE configuration. In case no database service is bound the fallback is the environment variable `DATABASE_URL`. 
+Using Cloud Foundry database services are detected from VCAP bindings and translated into Mendix M2EE configuration. In case no database service is bound the fallback is the environment variable `DATABASE_URL`.
 
-All database configuration code can be found in `lib\databaseconfig.py`. VCAP service bindings have preference over `DATABASE_URL`. VCAP are recognized on the identifier and tags.
+All database configuration code can be found in `lib/database_config.py`. VCAP service bindings have preference over `DATABASE_URL`. VCAP are recognized on the identifier and/or tags.
 
 ## Postgres using RDS Service Broker
 
-Selection based on tags `["database", "RDS", "postgresql"]`.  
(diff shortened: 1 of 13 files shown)


lib/database_config.py
```diff
@@ -0,0 +1,362 @@
...
+        return None
+
+
+class DatabaseConfiguration:
+    """Base clase for database configurations. Implements only the basics."""
+
+    def __init__(self):
+        development_mode = os.getenv("DEVELOPMENT_MODE", "").lower() == "true"
+
+    def get_database_configuration(self):
```
joshp123: Shouldn't it be called `get_m2ee_configuration` as per the docstring?

lib/database_config.py
```diff
@@ -0,0 +1,362 @@
...
+
+    database_type = "SAPHANA"
+
+    def __init__(self, credentials):
+        super().__init__()
+        logger.info("Detected SAP Hana configuration.")
+        self.credentials = credentials
+
+    def parse_configuration(self):
+        url = self.credentials["url"]
```
joshp123: You can make this simpler by just reusing the credentials dict as opposed to assigning all of these params to vars

lib/database_config.py
```diff
@@ -0,0 +1,362 @@
...
+        schema = self.credentials["schema"]
+        host = self.credentials["host"]
+        port = self.credentials["port"]
+        username = self.credentials["user"]
+        password = self.credentials["password"]
+
+        # parse parameters
```
joshp123: Handling query params and regular parameters (What's the difference?) seems like it can/should be done in a separate method.

docs/supported_database_vcaps.md
```diff
@@ -0,0 +1,70 @@
+# Overview of supported VCAP schemas
+
+Using Cloud Foundry database services are detected from VCAP bindings and translated into Mendix M2EE configuration. In case no database service is bound the fallback is the environment variable `DATABASE_URL`. 
+
+All database configuration code can be found in `lib\databaseconfig.py`. VCAP service bindings have preference over `DATABASE_URL`. VCAP are recognized on the identifier and tags.
```
svanderburg: `lib/databaseconfig.py` UNIX-style path name?

svanderburg:
With the exception of a small typo, it looks fine. Most likely, you also want to rebase this branch against master and resolve the merge conflict first.

PR #245, opened 2019-04-15
URLDecode database username

jpenninkhof:
The username passed in the URI parameter in VCAP_SERVICES may contain a @ character, which is URL encoded as %40. Hence, usernames should always be URL decoded. If this is working, the runtime will also support Cloud Platform scenario's, in which a connection to Azure PostgreSQL database is made using the Open Service Broker.

### DIFF
--- a/lib/database_config.py
+++ b/lib/database_config.py
@@ -4,7 +4,7 @@ import os
 import re
 import buildpackutil
 from abc import abstractmethod, ABC
-from urllib.parse import parse_qs, urlencode
+from urllib.parse import parse_qs, urlencode, unquote
 from m2ee import logger  # noqa: E402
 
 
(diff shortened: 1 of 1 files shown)


cwesdorp:
HI @jpenninkhof . Thanks for the merge request. Would it be possible that you send us a complete VCAP example for Azure Postgres via a support ticket? Then we can take a look if we can support it using the VCAP fields instead of parsing the uri.

cwesdorp:
@jpenninkhof I created a new branch decoding both username and password. Would it be possible for you to verify branch https://github.com/mendix/cf-mendix-buildpack/tree/urlencoded_jdbc_values?

cwesdorp:
Closing this one, is replaced by #276.

PR #247, opened 2019-04-18
Workaround for source push breakage after cflinuxfs3 upgrade

MXClyde:
- Since the upgrade to cflinuxfs3 source pushes fail.
- Root cause is that cflinuxfs3 does not contain libpng15 anymore (as libpng15 is deprecated from Bionic). To isolate this issue, run mxbuild with MONO_LOG_LEVEL=debug
- To work around the problem, this PR downloads a modified version of the mono archive that contains a forward-ported version of libpng15 (from Xenial). Obviously, these mono archives should be moved to cdn.mendix.com.


cwesdorp:
@MXClyde as discussed over internal communication we will handle this in the team.

PR #248, opened 2019-04-23
Add support for OpenShift

zlogic:
Switched to vanilla m2ee implementation of fix_permissions to avoid non-critical permission exceptions.


cwesdorp:
This merge request is replaced by https://github.com/mendix/cf-mendix-buildpack/pull/254

PR #253, opened 2019-05-16
double nginx worker connections

diogomatsubara:
This is being done to alleviate the 502 we're seeing. In our test environment this change helped and would like to give this a try. In the past (commit: cbefabdcb3ce96122a12dd39e23351a989f5be00) this was done for the exact same reason.

I'm not sure how's the release process for the cf-mendix-buildpack but I'd like to get this merged and deployed in the nonprod-1 environment and Free Apps cluster to verify how this change behaves under load.

### DIFF
--- a/nginx/conf/nginx.conf
+++ b/nginx/conf/nginx.conf
@@ -1,4 +1,4 @@
-worker_processes 1;
+worker_processes auto;
 daemon off;
 
 error_log stderr;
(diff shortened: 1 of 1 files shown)


diogomatsubara:
Our internal ticket is https://mendix.atlassian.net/browse/UP-1004

cwesdorp:
@diogomatsubara can you also update the version number in start.py please?

diogomatsubara:
Just out of completeness, after chatting with @cwesdorp IRL, we're going to test this change using one of our test apps in the prod environment before merging this into the master branch.

diogomatsubara:
Closed since we're not going to implement this for now. I'll re-open if later on we find out that we need this.

PR #257, opened 2019-06-17
DEP-616: print java version

cwesdorp:
With the increase variation of java versions (oracle, openjdk, 8, 11) let's print the java version on startup so when checking logs it becomes clear how the app is run.


svanderburg:
Looks good. Integrated.

PR #258, opened 2019-06-18
Disable the locale provider override for Java 11

jobvs:
For Mendix 8 with Java 11 we are switching back to the default value of the locale providers, so the override should only apply for Java 8.

Mx 6 and 7 => Java 8 => apply override
Mx 8 => Java 11 => do not apply override anymore

(DEP-1791)

### DIFF
--- a/start.py
+++ b/start.py
@@ -334,8 +334,8 @@ def get_constants(metadata):
 def set_jvm_locale(m2ee_section, java_version):
     javaopts = m2ee_section["javaopts"]
 
-    # enable locale for java8 or later
-    if not java_version.startswith("7"):
+    # override locale providers for java8
+    if java_version.startswith("8"):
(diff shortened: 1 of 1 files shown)


cwesdorp:
@jobvs  This change is not reflecting the PR title, this will change the locale setting for Java 8 and not Mendix 8. Can you please describe what the decision variables are in the PR and check the code?

KevinVlaanderen:
Effectively, the original code would override the local settings for Java 8 and higher, by checking that the Java version is NOT 7 (anything before 7 is considered irrelevant).
With this change, we explicitly override the locales for Java 8 only. For any other Java version (thus, 11), default locale settings will be used. For Java 11, this means that it will use "CLDR, SPI", which is what we want.

cwesdorp:
@svanderburg please merge this into mendix8beta3 branch

PR #261, opened 2019-07-01
Dep 1710 mono 5

svanderburg:


lib/buildpackutil.py
```diff
@@ -286,7 +286,9 @@ def _detect_mono_version(mx_version):
         "Detecting Mono Runtime using mendix version: " + str(mx_version)
     )
 
-    if mx_version < 7:
+    if mx_version >= 8:
```
cwesdorp: Lets make the check sequencial >= 8, >=7, else. Now it bounces from >=8 to < 7 and back to 7.

cwesdorp:
Looks good, will merge this. Test case passed (checked travis output)

PR #262, opened 2019-07-03
DVAS-326: Pushing sessions metrics for free apps

udh-mendix:
We are starting to push sessions metrics for free apps to influxDB. The following code changes are to support that.

### DIFF
--- a/lib/buildpackutil.py
+++ b/lib/buildpackutil.py
@@ -473,5 +473,5 @@ def get_metrics_url():
     return os.getenv("TRENDS_STORAGE_URL")
 
 
-def is_paid_app():
-    return os.getenv("PROFILE") != "free"
+def is_free_app():
+    return os.getenv("PROFILE") == "free"
--- a/lib/buildpackutil.py
+++ b/lib/buildpackutil.py
@@ -471,3 +471,7 @@ def bypass_loggregator_logging():
 
 def get_metrics_url():
     return os.getenv("TRENDS_STORAGE_URL")
+
+
+def is_paid_app():
+    return os.getenv("PROFILE") != "free"
--- a/lib/metrics.py
+++ b/lib/metrics.py
@@ -95,14 +95,18 @@ class MetricsEmitterThread(threading.Thread):
         while True:
             stats = {}
             try:
-                if buildpackutil.i_am_primary_instance():
-                    stats = self._inject_database_stats(stats)
-                    stats = self._inject_storage_stats(stats)
-                    stats = self._inject_health(stats)
                 try:
                     stats = self._inject_m2ee_stats(stats)
                 except Exception:
                     logger.debug("Unable to get metrics from runtime")
+
+                if (
+                    buildpackutil.i_am_primary_instance()
+                    and buildpackutil.is_paid_app()
+                ):
+                    stats = self._inject_database_stats(stats)
+                    stats = self._inject_storage_stats(stats)
+                    stats = self._inject_health(stats)
 
                 self.emit(stats)
             except psycopg2.OperationalError as up:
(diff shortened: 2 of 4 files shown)


lib/buildpackutil.py
```diff
@@ -471,3 +471,6 @@ def bypass_loggregator_logging():
 
 def get_metrics_url():
     return os.getenv("TRENDS_STORAGE_URL")
+
+def is_paid_app():
```
joshp123: I would do this the other way around for a couple of reasons:
- It is possible in principle to modify environment variables, so asking if an app is paid or not is a bit risky (customers could modify this) I think that it is safer to ask "is free app"
- The buildpack is used for applications outside of the Mendix cloud, so we want to design things in a way where restrictions (like this) are opt-in, as opposed to opt-out.

```suggestion
def is_free_app():
    return os.getenv("PROFILE") == "free"
```

lib/metrics.py
```diff
@@ -95,15 +95,16 @@ def run(self):
...
-                if buildpackutil.i_am_primary_instance():
-                    stats = self._inject_database_stats(stats)
-                    stats = self._inject_storage_stats(stats)
-                    stats = self._inject_health(stats)
                 try:
                     stats = self._inject_m2ee_stats(stats)
                 except Exception:
                     logger.debug("Unable to get metrics from runtime")
 
+                if buildpackutil.i_am_primary_instance() and buildpackutil.is_paid_app():
```
joshp123: Instead of gating this behind an `if` statement, I think that it's neater to subclass `MetricsEmitterThread` as a `FreeAppsEmitterThread`, and override the `run` method to call only the sessions stuff.

You can also put the logic to get specific metrics in their own internal `_get_free_apps_session_data` or whatever method.

start.py
```diff
@@ -1217,14 +1217,13 @@ def restart_callback():
 def start_metrics(m2ee):
     metrics_interval = os.getenv("METRICS_INTERVAL")
     profile = os.getenv("PROFILE")
-    if metrics_interval and profile != "free":
+    if metrics_interval:
```
joshp123: As above, I would change this block to initialize a `FreeAppsMetricsEmitterThread` is `buildpackutil.is_free_app` returns true.

start.py
```diff
@@ -1217,14 +1217,13 @@ def restart_callback():
...
         import metrics
-
         thread = metrics.MetricsEmitterThread(int(metrics_interval), m2ee)
         thread.setDaemon(True)
         thread.start()
     else:
-        logger.info("MENDIX-INTERNAL: Metrics are disabled.")
+        logger.info("MENDIX-INTERVAL not set: Metrics are disabled.")
```
joshp123: I wouldn't change this.

Also, the error message doesn't describe what's happening. `MENDIX-INTERVAL` is not what we are looking for, an d the message doesn't mention that this is an environment variable.

joshp123:
You will also want to make sure CI passes. It runs flake8 and Black, as well as other tests.

Please also add unit tests for the new functionality.

start.py
udh-mendix: the earlier log message also did not make sense. I would rather reword it and clarify that `MENDIX-INTERVAL` is now the environment variable that decides whether metrics get pushed or not.

lib/buildpackutil.py
udh-mendix: I thought of this and wasn't sure what method name would be better in the public-facing buildpack codebase. Will change.

start.py
joshp123: `MENDIX-INTERVAL` is _not_ the environment variable that decides this though. (it's `METRICS_INTERVAL`)

We (try to) prefix all non-debug log messages with `MENDIX-INTERNAL` to indicate that they are internal messages. Please leave this as is.

lib/buildpackutil.py
joshp123: It's less about naming, and more about gating functionality.

We don't want to _take away_ functionality, unless people opt in. Similarly, we don't want to _enable_ unexpected new functionality, unless people opt in.

We want to preserve things as they are for the most part, and use defaults that reflect this.

udh-mendix:
making changes as discussed. Closing this for now, will reopen when ready.

lib/metrics.py
udh-mendix: done. Changes in https://github.com/mendix/cf-mendix-buildpack/pull/262

PR #264, opened 2019-07-04
DVAS-326: Pushing sessions metrics for free apps

udh-mendix:
We are starting to push sessions metrics for free apps to influxDB. The following code changes are to support that.


lib/metrics.py
```diff
@@ -348,3 +348,41 @@ def _get_db_conn(self):
                 psycopg2.extensions.ISOLATION_LEVEL_AUTOCOMMIT
             )
         return self.db
+
+
+class FreeAppsMetricsEmitterThread(MetricsEmitterThread):
+    def run(self):
```
matyaskuti: What do you think of making this a template method in the base class? The structure is the same as in the `MetricsEmitterThread.run`, except for gathering stats. So by extracting the stats injection to its own method in the base class would allow to only override that and not having to copy the loop, sleeping and exception handling (which might cause problems in the future if we need to handle another exception or the current ones differently - we have to remember to change this in 2 places).

lib/metrics.py
```diff
@@ -348,3 +348,41 @@ def _get_db_conn(self):
...
+            stats = {}
+            try:
+                try:
+                    stats = self._inject_m2ee_stats(stats)
+                except Exception:
+                    logger.debug("Unable to get metrics from runtime")
+                # Only push session metrics for free apps
+                self.emit(stats["mendix_runtime"]["sessions"])
+            except psycopg2.OperationalError as up:
+                logger.exception("METRICS: error while gathering metrics")
```
matyaskuti: We could also log the stacktrace here if there is no specific reason not to. Also in L379.
```suggestion
                logger.exception("METRICS: error while gathering metrics"., exc_info=True)
```
@cwesdorp would that cause any issues?

matyaskuti: Never mind, `logger.exception` already takes car of this: https://docs.python.org/3/library/logging.html#logging.Logger.exception

lib/metrics.py
pl-jankowskimichal: As suggested, I have change MetricsEmitterThread to BaseMetricsEmitterThread, form which FreeAppsMetricsEmitterThread and PaidAppsMetricsEmitterThread inherit.

lib/metrics.py
```diff
@@ -348,3 +348,41 @@ def _get_db_conn(self):
...
+        logger.debug(
+            "Starting free apps metrics emitter with interval %d"
+            % self.interval
+        )
+        while True:
+            stats = {}
+            try:
+                try:
+                    stats = self._inject_m2ee_stats(stats)
+                except Exception:
```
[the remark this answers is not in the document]
pl-jankowskimichal: I removed redundant try-except clause.

Makefile
```diff
@@ -3,4 +3,4 @@ test: lint
 
 lint:
 	flake8 --exclude .git,__pycache__,lib/certifi,lib/idna,lib/psycopg2,lib/urllib3,lib/chardet,lib/httplib2,lib/requests,lib/yaml
-	black --check --diff .
+	# black --check --diff .
```
joshp123: a) Commented code
b) If we're disabling black because of the python version changes, can we figure out how to get Black working under e.g. python 3.6, whilst still running our tests with 3.4?

lib/metrics.py
```diff
@@ -348,3 +358,33 @@ def _get_db_conn(self):
...
+            ]
+        selected_stats.append(self._inject_m2ee_stats)
+        return selected_stats
+
+
+class FreeAppsMetricsEmitterThread(BaseMetricsEmitterThread):
+    def _extract_session_metrics(self, stats):
+
+        try:
+            session_metrics = stats["mendix_runtime"]["sessions"]
```
joshp123: Instead of stripping out metrics after the fact, wouldn't it be simpler to only emit the stats that actually need emitting by overriding the run method?

lib/metrics.py
```diff
@@ -82,49 +87,54 @@ def __init__(self, interval, m2ee):
...
+    def _set_stats_info(stats):
         stats["version"] = "1.0"
         stats["timestamp"] = datetime.datetime.now().isoformat()
         stats["instance_index"] = os.getenv("CF_INSTANCE_INDEX", 0)
+        return stats
+
+    def emit(self, stats):
         self.emitter.emit(stats)
 
+    @abstractmethod
```
joshp123: I would make this an `abstractproperty` instead of an abstract method, and make the subclass implementations properties too.

tests/usecase/test_emit_metrics.py
```diff
@@ -15,6 +15,22 @@ def test_read_metrics_in_logs(self):
...
+            "sample-6.2.0.mda",
+            env_vars={
+                "METRICS_INTERVAL": "10",
+                "PROFILE": "free",
+            },
+        )
+        self.startApp()
+
+        time.sleep(10)
+        self.assert_string_in_recent_logs("MENDIX-METRICS: ")
```
joshp123: This relies on the fact that metrics are emitted via stdout when `BYPASS_LOGGREGATOR` and the trends-storage-server URL environment variables are both set. 

In production, we won't be emitting metrics over stdout, so this test doesn't really accurately test the code path that we want.

(This probably isn't a big deal, but I think that this discrepancy should at least be explained in the docstring of this method)

tests/usecase/test_sap_hana.py
```diff
@@ -1,5 +1,7 @@
 import json
 import os
+import urllib.parse
```
joshp123: Why are we changing SAP/Hana code in a PR that doesn't relate to SAP/Hana?

matyaskuti: These tests were flaky, relying on dictionary ordering (which they shouldn't), revealed by pinning the Python version to 3.4.3

lib/metrics.py
```diff
@@ -67,9 +67,14 @@ def emit(self, stats):
...
-class MetricsEmitterThread(threading.Thread):
+class BaseMetricsEmitterThread(threading.Thread, metaclass=ABCMeta):
+    """
+    This base class contains all boilerplate code needed to emit metrics.
+    Only thing left to implement in subclass is `_selected_stats_to_emit` method.
+    """
+
     def __init__(self, interval, m2ee):
-        super(MetricsEmitterThread, self).__init__()
+        super(BaseMetricsEmitterThread, self).__init__()
```
matyaskuti:
```suggestion
        super().__init__()
```
Should do the trick, since we are using Python3.

Makefile
pl-jankowskimichal: this is a temporary situation. IMHO we should ASAP migrate to a higher version of python. 

I'm not interested in creating bandaid, but if you want to check black under 3.6 and run code under 3.4 than go ahead :)

lib/metrics.py
pl-jankowskimichal: Do you mean making changes to m2ee module?  This method is executed before emmiting metrics not after.
All runtime metrics are fetched via `munin.get_stats_from_runtime` in` _inject_m2ee_stats()` then only session metrics are extracted with this method. Then metrics get emmited.

lib/metrics.py
pl-jankowskimichal: Cloud you elaborate why?

joshp123: https://stackoverflow.com/questions/20506612/python-property-vs-method-when-no-access-to-attribute-is-needed

tl;dr returning simple values that are essentially static/constant for an instance of the class, it's just more elegant.

tests/usecase/test_sap_hana.py
joshp123: Okay, cool.

Let's make sure that someone who is familiar with the code (@cwesdorp going from git blame) has reviewed this before we merge, since these test changes seem functional (i.e. supplied URLs are checked differently).

Makefile
joshp123: I also think that we should migrate to a higher version of Python, but that's not something that our team has control over.

I think that we should indeed perform black checks under python 3.6, and run the code under python 3.4.

Some quick googling found this link, which suggests that this is fairly trivial: https://stackoverflow.com/a/20621143/2230226

Can you please implement this? I really don't want us to lose black in the buildpack, as I think that it adds a lot of readability, and enforcing this in CI is a fairly elegant way to do things.

matyaskuti: It might be worth looking into the build stages functionality of Travis: https://docs.travis-ci.com/user/build-stages/

lib/metrics.py
joshp123: No, I mean something a bit different:

1) We call `BaseMetricsEmitterThread._inject_m2ee_stats`, which itself calls out to `munin`, `m2ee.client.get_critical_log_messages`, and a few other things. But we only want a subset of those things. I think that it would be better to simply add a `_get_user_sessions_stats` method to get the user sessions stats, and not, for example, the critical log messages count. 

    If you wanted to call this new method from `_inject_m2ee_stats`_, that would be fine.

2) In `BaseMetricsEmitterThread.run,  we loop over all selected stats methods to emit, and then call those methods. Depending on if there were exceptions thrown during these methods, we then add health information; either good or bad. However, health information isn't relevant for free apps. I think that the try/except calls should be different for free apps and for paid apps. For paid apps, they should remain the same. For free apps, we should simply do nothing in the case of an exception.

Makefile
joshp123: That looks even better!

matyaskuti: PR into this one: https://github.com/mendix/cf-mendix-buildpack/pull/267

lib/metrics.py
```diff
@@ -348,3 +358,33 @@ def _get_db_conn(self):
...
+        except KeyError:
+            session_metrics = {}
+        return session_metrics
+
+    def _selected_stats_to_emit(self):
+        """
+        For free apps metrics we only care about sessions metrics.
+        We use `self._extract_session_metrics` to shrink original stats object before emitting.
+        """
+        return [self._inject_m2ee_stats, self._extract_session_metrics]
```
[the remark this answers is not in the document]
pl-jankowskimichal: You are raising a good point here.

lib/metrics.py
```diff
@@ -82,49 +87,54 @@ def __init__(self, interval, m2ee):
...
-                self.emit(
-                    {
-                        "health": {
-                            "health": 0,
-                            "diagnosis": "Database error: %s" % str(up),
-                        }
+                stats = {
+                    "health": {
+                        "health": 0,
+                        "diagnosis": "Database error: %s" % str(up),
```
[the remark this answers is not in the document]
pl-jankowskimichal: In my opinion, such changes should not be a part of this PR, since they are out of scope. 
However, I will refactor run method anyway, so why not..

lib/metrics.py
pl-jankowskimichal: Well. I wanted to not touch the original code as much as possible. But since we are on it.

1. Maybe you're right here, also after taking into account what Chris mentioned about nasty side effects if `extract_session_metrics` is the last called method. I will create `_get_user_sessions_stats` method instead using _inject_m2ee_stats_ followed by `_extract_session_metrics`. 

2. In my opinion, all the _inject methods should catch their own exceptions and handle them accordingly so that there should be no need for "try exception" clause in the run method at all. `run` should just be a dummy wrapper, no logic, no test required. imho.
The fact that we degrade the health status of the application just because there was ANY exception during gathering metrics is weird. Also, any successfully gathered metrics are lost due to the fact that we are simply replacing the stats dictionary, instead of replacing the value(s) under "health" key

All of those changes mean that I should introduce more tests to the project.

tests/usecase/test_emit_metrics.py
pl-jankowskimichal: I left the test as it was since it was in sync with the style of the tests of this project. But since I'am refactoring more and more, it should be easier (and more desired) to introduce small scoped unit tests

lib/metrics.py
joshp123: I strongly agree with not touching the original code as much as possible.

I think that for the sake of simplicity, you can create your own simpler `.run` method on the `FreeAppsMetricEmitter` class, and move the `.run` method on `BaseAppsMetricEmitter` to `PaidAppsMetricEmitter`, but leaving the code as it was before your changes.

This means we can keep free apps elegant, but without changing anything for existing apps.

Re: point 2, I agree that this is weird (indeed, this is the source of some annoying problems with alerting). But, like your first point, I would again change this as little as is possible.

joshp123: >I will create _get_user_sessions_stats method instead using inject_m2ee_stats followed by _extract_session_metrics.

I would do this differently:

```python
    def _inject_free_apps_session_stats(self, stats):
        m2ee_stats, _ = munin.get_stats_from_runtime(
            self.m2ee.client, self.m2ee.config
        )
        stats["mendix_runtime"] = {}
        stats["mendix_runtime"]["sessions"] = m2ee_stats["sessions"]
        return stats

    # This is unchanged from the current situation
    def _inject_m2ee_stats(self, stats):
        m2ee_stats, java_version = munin.get_stats_from_runtime(
            self.m2ee.client, self.m2ee.config
        )
        if "sessions" in m2ee_stats:
            m2ee_stats["sessions"]["user_sessions"] = {}
        m2ee_stats = munin.augment_and_fix_stats(
            m2ee_stats, self.m2ee.runner.get_pid(), java_version
        )

        critical_logs_count = len(self.m2ee.client.get_critical_log_messages())
        m2ee_stats["critical_logs_count"] = critical_logs_count
        stats["mendix_runtime"] = m2ee_stats
        return stats
```

This would cut down on the number of m2ee calls for free apps, which would IMO be preferable.

pl-jankowskimichal: How is that different from
>I will create _get_user_sessions_stats method instead using inject_m2ee_stats followed by _extract_session_metrics. 
??
Also why don't you just commit this to this PR? This discussion is going to some weird place.

tests/usecase/test_emit_metrics.py
```diff
@@ -79,3 +108,54 @@ def test_posting_metrics_works(self):
...
+        self.metrics_emitter.emit = Mock()
+        self.metrics_emitter.setDaemon(True)
+
+    def test_inject_user_session_metrics(self):
+        stats = {"key": "value"}
+        expected_stats = copy.deepcopy(stats)
+        expected_stats["mendix_runtime"] = mock_user_session_metrics
+
+        stats = self.metrics_emitter._inject_user_session_metrics(stats)
+        self.assertEqual(True, self.metrics_emitter._get_munin_stats.called)
```
matyaskuti:
```suggestion
        self.assertTrue(self.metrics_emitter._get_munin_stats.called)
```

Please use `self.assertTrue` instead. Semantically it doesn't change anything, but reads nicer and documents the expected behaviour better.

While there is a valid usecase of the `self.assertEqual(True, ...)` structure - in case we want to make sure the returned value is `True`, explicitly, instead of just truthy which `self.assertTrue` checks - this is not one of them due to the `Mock.called` behaviour.

tests/usecase/test_emit_metrics.py
```diff
@@ -79,3 +108,54 @@ def test_posting_metrics_works(self):
...
+    ):
+        stats = {
+            "key": "value",
+            "mendix_runtime": {"other_key": "other_value"},
+        }
+        expected_stats = copy.deepcopy(stats)
+        expected_stats["mendix_runtime"].update(mock_user_session_metrics)
+
+        stats = self.metrics_emitter._inject_user_session_metrics(stats)
+        self.assertEqual(True, self.metrics_emitter._get_munin_stats.called)
```
matyaskuti:
```suggestion
        self.assertTrue(self.metrics_emitter._get_munin_stats.called)
```

Please use `self.assertTrue` instead.

tests/usecase/test_emit_metrics.py
```diff
@@ -79,3 +108,54 @@ def test_posting_metrics_works(self):
...
+    def test_inject_user_session_metrics_when_exception_raised(self):
+        stats = {"key": "value"}
+        expected_stats = copy.deepcopy(stats)
+
+        with self.assertRaises(Exception):
+            self.metrics_emitter._get_munin_stats = Mock(
+                side_effect=Exception("M2EE Exception!")
+            )
+            stats = self.metrics_emitter._inject_user_session_metrics(stats)
+            self.assertEqual(
```
matyaskuti: Please use `self.assertTrue` instead.

tests/usecase/test_emit_metrics.py
```diff
@@ -79,3 +108,54 @@ def test_posting_metrics_works(self):
...
+
+    def test_inject_user_session_metrics(self):
+        stats = {"key": "value"}
+        expected_stats = copy.deepcopy(stats)
+        expected_stats["mendix_runtime"] = mock_user_session_metrics
+
+        stats = self.metrics_emitter._inject_user_session_metrics(stats)
+        self.assertEqual(True, self.metrics_emitter._get_munin_stats.called)
+        self.assertEqual(expected_stats, stats)
+        print(stats)
```
matyaskuti:
```suggestion
```

Leftover `print`, please remove

lib/metrics.py
```diff
@@ -348,3 +346,77 @@ def _get_db_conn(self):
...
+        selected_stats.append(self._inject_m2ee_stats)
+        return selected_stats
+
+    def _gather_metrics(self):
+        stats = {}
+        try:
+            for inject_method in self._select_stats_to_emit:
+                stats = inject_method(stats)
+        except psycopg2.OperationalError as up:
+            logger.exception("METRICS: error while gathering metrics")
```
matyaskuti: What do you think of extending this error message with a bit more info relating the issue to the database?
I know that line numbers show up in the exception and thus in the logs so we can pinpoint where it came from, but it might be easier to distinguish this from the catchall clause in L377 if we use a more informative error message.

lib/metrics.py
```diff
@@ -348,3 +346,77 @@ def _get_db_conn(self):
...
+            ]
+        selected_stats.append(self._inject_m2ee_stats)
+        return selected_stats
+
+    def _gather_metrics(self):
+        stats = {}
+        try:
+            for inject_method in self._select_stats_to_emit:
+                stats = inject_method(stats)
+        except psycopg2.OperationalError as up:
```
matyaskuti:
```suggestion
        except psycopg2.OperationalError as exc:
```

I think the capture exception name `up` is somewhat confusing, especially further down in this block. Could you change it to the "usual" `exc` in accordance with the principle of least astonishment?

lib/metrics.py
```diff
@@ -348,3 +346,77 @@ def _get_db_conn(self):
...
+            logger.exception("METRICS: error while gathering metrics")
+            stats = {
+                "health": {
+                    "health": 0,
+                    "diagnosis": "Database error: {}".format(str(up)),
```
matyaskuti:
```suggestion
                    "diagnosis": "Database error: {}".format(str(exc)),
```

See comment on L369.

lib/metrics.py
```diff
@@ -67,9 +67,14 @@ def emit(self, stats):
             self.fallback_emitter.emit(stats)
 
 
-class MetricsEmitterThread(threading.Thread):
+class BaseMetricsEmitterThread(threading.Thread, metaclass=ABCMeta):
+    """
+    This base class contains all boilerplate code needed to emit metrics.
+    Only thing left to implement in subclass is `_selected_stats_to_emit` method.
```
matyaskuti: Could you update this dosctring? Since `_selected_stats_to_emit` is a property now and there is another abstract method `_gather_metrics`.

lib/metrics.py
joshp123: Bear in mind that these all buildpack error messages show up in customer logs, and they result in tickets, because customers check logs, and open support tickets when there are errors.

Often we cannot do much about those errors.

We should be extremely cautious when modifying or adding extra buildpack logging, as it ends up in customer logs.

(The ideal situation is that we have our own private logs for internal buildpack logging, but that won't happen particularly soon)

lib/metrics.py
```diff
@@ -348,3 +346,77 @@ def _get_db_conn(self):
...
+                self._inject_storage_stats,
+                self._inject_health,
+            ]
+        selected_stats.append(self._inject_m2ee_stats)
+        return selected_stats
+
+    def _gather_metrics(self):
+        stats = {}
+        try:
+            for inject_method in self._select_stats_to_emit:
```
joshp123: This causes a subtle but important change from the previous functionality.

Because we added the `_select_stats_to_emit` method, now, if there is an error during `_inject_m2ee_stats`, we will mark the application as unhealthy (line 381), and also log an exception.

This is different from the previous behaviour: https://github.com/mendix/cf-mendix-buildpack/pull/264/files#diff-dd78eb8c4863775411e7a54feb3946b9L104
Previously if there was an error whilst capturing m2ee metrics, we do _not_ mark the application  as unhealthy, and we do _not_ log anything (unless debug mode was on).

We absolutely need to preserve this behaviour, as it was added in response to customer support tickets, and was a major contributor to false positive alerts.

lib/metrics.py
```diff
@@ -348,3 +346,77 @@ def _get_db_conn(self):
...
+        return m2ee_stats
+
+    def _inject_user_session_metrics(self, stats):
+        session_metrics = {}
+        try:
+            m2ee_stats = self._get_munin_stats()
+            if "sessions" in m2ee_stats.keys():
+                m2ee_stats["sessions"]["user_sessions"] = {}
+                session_metrics = m2ee_stats["sessions"]
+        except Exception:
```
joshp123: As mentioned above, we should be very careful when introducing extra buildpack logging, as it is exposed to customers.

In this case, I think that we do not need to log anything here. If runtime metrics aren't emitted, it's not worth putting into a customer's logs. (Even for a free app).

lib/metrics.py
pl-jankowskimichal: According to docs logger.exception already includes Exception info, so I don't think modifying log message is necessary. https://docs.python.org/3/library/logging.html#logging.exception

Please also notice that while I moved this logic from run to separate method don't want to change anything about the logic. 

As I have mentioned in some other discussion, for me personally it is weird that we decided to not include all other available metrics in case of exception while gathering a single category.

We already have broadened up the scope of this PR, outside of the original one, I don't think proper logging for Paid Apps falls under scope of this PR

What do you think?

lib/metrics.py
pl-jankowskimichal: That is true, I think I wanted to put that try exception clause inside _inject_m2ee_stats, and forgot about it. Let me fix it right away.

joshp123: Yes, ensuring that `_inject_m2ee_stats` would never throw an exception would do the trick.

lib/metrics.py
pl-jankowskimichal: Please check now.

tests/usecase/test_emit_metrics.py
pl-jankowskimichal: Changed!

tests/usecase/test_emit_metrics.py
```diff
@@ -79,3 +108,51 @@ def test_posting_metrics_works(self):
...
+    def test_inject_user_session_metrics_when_exception_raised(self):
+        stats = {"key": "value"}
+        expected_stats = copy.deepcopy(stats)
+
+        with self.assertRaises(Exception):
+            self.metrics_emitter._get_munin_stats = Mock(
+                side_effect=Exception("M2EE Exception!")
+            )
+            stats = self.metrics_emitter._inject_user_session_metrics(stats)
+            self.assertTrue(self.metrics_emitter._get_munin_stats.called)
```
matyaskuti: If any of the above lines in this `with` block raise an exception, then these asserts in L157-158 will not be invoked.

lib/metrics.py
pl-jankowskimichal: Aren't we interested in reasons for failure? Since all logs go through CF Loggregator we can easily get insight on the root reason(s) for runtime metrics not being fetched/emitted. :thinking:

tests/usecase/test_emit_metrics.py
pl-jankowskimichal: right self. assertTrue and self.assertEqual should be outside of the context manager.

lib/metrics.py
pl-jankowskimichal: Done

lib/metrics.py
joshp123: No, we are not interested.

This is because the reasons are usually performance of customer apps, which is outside of our control.

They will do CPU/memory intensive stuff as part of their application, which will cause the runtime not to respond. This is normal behaviour. It's is not our responsibility (as operators of the Mendix cloud), but the responsibility of the customer (as the operator of their application).

We do not want to log anything regarding a _symptom_ of the problem (e.g. a timeout talking to the runtime), but instead want to expose stability problems via our usual mechanisms (i.e. alerts)

joshp123: This also has a heavy cost for us too, as we end up investigating these issues (in response to support tickets) and concluding that it is a performance problem for which the customer is responsible.

joshp123: As I mention above, the ideal situation here is something like Sentry where we can privately aggregate such errors in the buildpack and investigate if we deem necessary. But we do not have such a facility now, and won't have it any time soon.

pl-jankowskimichal: As discussed in person I have lowered the log level for those messages. In case internal log aggregator fo customer apps are in place we can still utilize those log messages, and they still stay 'hidden' from Free Apps cluster end users.

tests/usecase/test_emit_metrics.py
```diff
@@ -79,3 +108,50 @@ def test_posting_metrics_works(self):
...
+        expected_stats = copy.deepcopy(stats)
+        expected_stats["mendix_runtime"].update(mock_user_session_metrics)
+
+        stats = self.metrics_emitter._inject_user_session_metrics(stats)
+        self.assertTrue(self.metrics_emitter._get_munin_stats.called)
+        self.assertEqual(expected_stats, stats)
+
+    def test_inject_user_session_metrics_when_exception_raised(self):
+        stats = {"key": "value"}
+        expected_stats = copy.deepcopy(stats)
```
matyaskuti: According to the latest pipeline run, `expected_stats` is missing `"sessions": {}`

pl-jankowskimichal: right, session metrics will always be injected even if it's an empty object.

pl-jankowskimichal:
@svanderburg @cwesdorp could you guys review this PR?

lib/metrics.py
```diff
@@ -348,3 +352,79 @@ def _get_db_conn(self):
...
+        m2ee_stats, _ = munin.get_stats_from_runtime(
+            self.m2ee.client, self.m2ee.config
+        )
+        return m2ee_stats
+
+    def _inject_user_session_metrics(self, stats):
+        session_metrics = {}
+        try:
+            m2ee_stats = self._get_munin_stats()
+            if "sessions" in m2ee_stats.keys():
```
joshp123: You don't need to call the `keys()` function here. You can just use `if x in y`:
```
>>> a  = {'foo': 123, 'bar': 456}
>>> 'foo' in a
True
>>> 'baz' in a
False
>>>
```

```suggestion
            if "sessions" in m2ee_stats:
```

lib/metrics.py
```diff
@@ -348,3 +352,79 @@ def _get_db_conn(self):
...
+                session_metrics = m2ee_stats["sessions"]
+        except Exception:
+            logger.debug(
+                "METRICS: error while gathering runtime metrics", exc_info=True
+            )
+        finally:
+            # runtime metrics shouldn't be included,
+            # but if they were due to some upstream code changes,
+            # we would ensure to not override all of them.
+            if "mendix_runtime" not in stats.keys():
```
joshp123:
```suggestion
            if "mendix_runtime" not in stats:
```

tests/run.sh
```diff
@@ -25,6 +25,6 @@ cleanup
 echo 'starting test run, tests will run in parallel and output shown at the end'
 
 export PYTHONPATH=$PWD/..:$PWD/../lib/
-nosetests --verbosity=3 --processes=10 --process-timeout=3600 --with-timer --timer-no-color usecase/test_*.py
```
joshp123: How come we don't want the verbosity or the timers?

tests/usecase/test_emit_metrics.py
joshp123: Can you please still add the above as a docstring on the `test_free_apps_metrics` method?

tests/usecase/test_emit_metrics.py
```diff
@@ -15,6 +21,29 @@ def test_read_metrics_in_logs(self):
...
+        self.startApp()
+
+        time.sleep(10)
+        self.assert_string_in_recent_logs("MENDIX-METRICS: ")
+        self.assert_string_in_recent_logs("named_users")
+        self.assert_string_in_recent_logs("anonymous_sessions")
+        self.assert_string_in_recent_logs("named_user_sessions")
+
+
+mock_user_session_metrics = {
```
joshp123: `mock_user_session_metrics` is defined on line 38, but the class that uses it is defined on line 111.

Can we a) move it closer to the class that uses it and b) consider making it a member variable of the class that uses it (only one class uses it)

lib/headers.py
```diff
@@ -0,0 +1,58 @@
+import json
+import os
+import re
+
+from m2ee import logger  # noqa: E402
```
joshp123: We don't actually need this `noqa: E402`, (E402 module level import not at top of file), as the module level imports _are_ at the top of the file.

tests/run.sh
pl-jankowskimichal: 1) Output got truncated anyway
2) From the conversations we had, It was clear no one checks or cares about the execution time and/or verbose logs.
3) We got rid of the dependency

cwesdorp: Let's try to get timing back now test run okay, having a overview of all tests run if useful.

matyaskuti:
@joshp123 @cwesdorp @svanderburg Comments resolved, please take a look again.

tests/usecase/test_emit_metrics.py
```diff
@@ -1,8 +1,25 @@
 import basetest
+import copy
 import time
 
+from unittest import TestCase
+from unittest.mock import Mock
+
+from metrics import FreeAppsMetricsEmitterThread
```
joshp123: shouldn't this actually be `from lib.metrics import blah`?

I suspect that this only works as a side-effect of some of the horrible path manipulation things

matyaskuti: True, it's horrible path manipulation on all levels. This is exactly why I wouldn't break the pattern now because that might have some unforseen consequences that could take a long and tedious time to fix, or just simply adding more to the "snowflakiness".

tests/usecase/test_emit_metrics.py
```diff
@@ -15,6 +32,19 @@ def test_read_metrics_in_logs(self):
         self.assert_string_in_recent_logs("number_of_files")
         self.assert_string_in_recent_logs("critical_logs_count")
 
+    def test_free_apps_metrics(self):
```
udh-mendix: I dont think we need this method anymore, now that we have `TestFreeAppsMetricsEmitter`

matyaskuti: `TestFreeAppsMetricsEmitter` has more fine-grained unit tests as opposed to this integration test. I think it's actually better to have both.

matyaskuti:
@svanderburg As discussed, I bumped the version number.

PR #265, opened 2019-07-11
Experiment with different cf stacks

cwesdorp:


cwesdorp:
Will pick this up again later

PR #270, opened 2019-07-30
support vcap, database url or custom runtime settings for the database config

cwesdorp:
Updates the logic that determines which database configuration to use. The issue with the current implementation was that it didn't fall back on MXRUNTIME_Database.... values causing that only the database types explicitly coded in the buildpack where supported either via a VCAP or DATABASE_URL environment variable. 

Change makes it also possible for the docker buildpack to use either DATABASE_URL or custom runtime settings.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,7 +54,12 @@ In our trial we found the service `elephantsql` which offered the free `turtle` 
 
     cf bind-service <YOUR_APP> <SERVICE_NAME>
 
-Note that not all databases are automatically picked up by the buildpack. If `cf push` returns an error like `Could not parse database credentials`, you need to set the `DATABASE_URL` variable manually or set database [Mendix custom runtime variables](https://docs.mendix.com/refguide/custom-settings) to configure a database. Note these need to be prefixed with `MXRUNTIME_`. 
+Note that not all databases are automatically picked up by the buildpack. If `cf push` returns an error like `Could not parse database credentials`, you need to set the `DATABASE_URL` variable manually or set database [Mendix custom runtime variables](https://docs.mendix.com/refguide/custom-settings) to configure a database. Note these variables need to be prefixed with `MXRUNTIME_`, as per example:
+
+    cf set-env <YOUR_APP> MXRUNTIME_DatabaseType PostgreSQL
+    cf set-env <YOUR_APP> MXRUNTIME_DatabaseJdbcUrl postgres://host/databasename
+    cf set-env <YOUR_APP> MXRUNTIME_DatabaseUsername user
+    cf set-env <YOUR_APP> MXRUNTIME_DatabasePassword password
 
 Now we need to push the application once more.
 
@@ -405,8 +410,8 @@ cf restart <YOUR_APP>
 Contributing
 ====
 
-Make sure your code complies with PEP8. Make sure your code is styled using [Black](https://github.com/psf/black). 
-We enforce this using `flake8` and `black` in our travis CI. 
+Make sure your code complies with PEP8. Make sure your code is styled using [Black](https://github.com/psf/black).
+We enforce this using `flake8` and `black` in our travis CI.
 
 This simplest way to use these tools is by installing them as a plugin for
 your editor; for example in Vim, one can auto-format files with `black` on writing out a buffer, and it will also display `flake8` errors.
--- a/README.md
+++ b/README.md
@@ -54,7 +54,7 @@ In our trial we found the service `elephantsql` which offered the free `turtle` 
 
     cf bind-service <YOUR_APP> <SERVICE_NAME>
 
-Note that not all databases are automatically picked up by the buildpack. If `cf push` returns an error like `Could not parse database credentials`, you need to set the `DATABASE_URL` variable manually using the details included in the service.
+Note that not all databases are automatically picked up by the buildpack. If `cf push` returns an error like `Could not parse database credentials`, you need to set the `DATABASE_URL` variable manually or set database [Mendix custom runtime variables](https://docs.mendix.com/refguide/custom-settings) to configure a database. Note these need to be prefixed with `MXRUNTIME_`. 
 
 Now we need to push the application once more.
 
--- a/lib/database_config.py
+++ b/lib/database_config.py
@@ -58,10 +58,10 @@ class DatabaseConfigurationFactory:
 
         # fallback to original configuration
         url = self.get_database_uri_from_vcap(self.vcap_services)
-        if url is None and "DATABASE_URL" in os.environ:
+        if not url and "DATABASE_URL" in os.environ:
             url = os.environ["DATABASE_URL"]
 
-        if url is not None:
+        if url:
             return UrlDatabaseConfiguration(url)
 
         return None
--- a/lib/database_config.py
+++ b/lib/database_config.py
@@ -9,13 +9,15 @@ from m2ee import logger  # noqa: E402
 
 
 def get_database_config(development_mode=False):
-    # the following options are validated to get database credentials
-    # 1) existence of custom runtime settings Database.... values
-    # 2) VCAP with database credentials
-    # 3) existence of DATABASE_URL
-    #
-    # In case we find MXRUNTIME_Database.... values we don't interfere and
-    # return nothing. VCAP or DATABASE_URL return m2ee configuration
+    """
+    the following options are validated to get database credentials
+    1) existence of custom runtime settings Database.... values
+    2) VCAP with database credentials
+    3) existence of DATABASE_URL env var
+
+    In case we find MXRUNTIME_Database.... values we don't interfere and
+    return nothing. VCAP or DATABASE_URL return m2ee configuration
+    """
     if any(
         [x.startswith("MXRUNTIME_Database") for x in list(os.environ.keys())]
     ):
@@ -29,10 +31,10 @@ def get_database_config(development_mode=False):
 
     if configuration:
         m2ee_config = configuration.get_m2ee_configuration()
-        if m2ee_config is not None and "DatabaseType" in m2ee_config:
+        if m2ee_config and "DatabaseType" in m2ee_config:
             return m2ee_config
 
-    raise Exception(
+    raise RuntimeError(
         "Can't find database configuration from environment variables. "
         "Check README for supported configuration options."
     )
--- a/lib/database_config.py
+++ b/lib/database_config.py
@@ -9,15 +9,33 @@ from m2ee import logger  # noqa: E402
 
 
 def get_database_config(development_mode=False):
+    # the following options are validated to get database credentials
+    # 1) existence of custom runtime settings Database.... values
+    # 2) VCAP with database credentials
+    # 3) existence of DATABASE_URL
+    #
+    # In case we find MXRUNTIME_Database.... values we don't interfere and
+    # return nothing. VCAP or DATABASE_URL return m2ee configuration
     if any(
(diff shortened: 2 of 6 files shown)


README.md
```diff
@@ -54,7 +54,7 @@ In our trial we found the service `elephantsql` which offered the free `turtle`
 
     cf bind-service <YOUR_APP> <SERVICE_NAME>
 
-Note that not all databases are automatically picked up by the buildpack. If `cf push` returns an error like `Could not parse database credentials`, you need to set the `DATABASE_URL` variable manually using the details included in the service.
+Note that not all databases are automatically picked up by the buildpack. If `cf push` returns an error like `Could not parse database credentials`, you need to set the `DATABASE_URL` variable manually or set database [Mendix custom runtime variables](https://docs.mendix.com/refguide/custom-settings) to configure a database. Note these need to be prefixed with `MXRUNTIME_`.
```
joshp123: Note these need -> Note that these variables need 

or set database Mendix custom runtime variables to configure a database -> or set 

I would add an example of prefixing the variables also, i.e `Database_Foo: bar` -> `MXRUNTIME_Database_Foo: bar`

lib/database_config.py
```diff
@@ -9,15 +9,33 @@
 
 
 def get_database_config(development_mode=False):
+    # the following options are validated to get database credentials
```
joshp123: This is a docstring not a comment, so use triple-quote syntax.

I would also say:

To get database credentials, we check and validate these options in order:
...

We _could_ also mention that VCAP and DATABASE_URL are env vars, because this might not be clear to a newcomer

lib/database_config.py
```diff
@@ -9,15 +9,33 @@
...
+        )
         return None
 
     factory = DatabaseConfigurationFactory()
     configuration = factory.get_instance()
 
-    return configuration.get_m2ee_configuration()
+    if configuration:
+        m2ee_config = configuration.get_m2ee_configuration()
+        if m2ee_config is not None and "DatabaseType" in m2ee_config:
```
joshp123: Do you want `if m2ee_config is not None` (i.e. `[]` (an empty list) would return `True` here) or do you actually want `if m2ee_config`? (only "truthy" variables will return `True`)

lib/database_config.py
```diff
@@ -9,15 +9,33 @@
...
+            return m2ee_config
+
+    raise Exception(
```
joshp123: It's probably better to raise a `RuntimeError` as opposed to a raw `Exception`.

lib/metrics.py
```diff
@@ -327,12 +328,19 @@ def _get_db_conn(self):
             self.db = None
 
         if not self.db:
+            # get_database config may return None or empty
             db_config = database_config.get_database_config()
+            if db_config is None or "DatabaseType" not in db_config:
```
joshp123: As above, you probably want `if not db_config` here.

tests/usecase/test_db_config_options.py
```diff
@@ -0,0 +1,85 @@
...
+        """
+        Setting different environment variables for test in the same process
+        can lead to flaky tests.
+        """
+        if "DATABASE_URL" in os.environ.keys():
+            del os.environ["DATABASE_URL"]
+
+        for key in filter(
+            lambda x: x.startswith("MXRUNTIME_Database"),
+            list(os.environ.keys()),
```
matyaskuti: No need for the `list(...)` conversion, and the `.keys()` call, `os.environ` is already iterable by the keys.
Alternatively to filter you could also use a generator expression: `(k for k in os.environ if k.startswith("..."))`

tests/usecase/test_db_config_options.py
```diff
@@ -0,0 +1,85 @@
...
+        ):
+            del os.environ[key]
+
+    def test_no_setup(self):
+        self.clean_env()
+        try:
+            config = get_database_config()
```
matyaskuti: If I understand correctly, in this case this call is supposed to raise an exception, meaning the `assert config is None` will never be called.
Furthermore, could you use a `with self.assertRaises(RuntimeError)` (used RuntimeError in the example as per @joshp123's comment) context manager to assert the exception was raised?

tests/usecase/test_db_config_options.py
```diff
@@ -0,0 +1,85 @@
...
+            assert config is None
+        except Exception as e:
+            assert "Can't find database configuration" in str(e)
```
matyaskuti: Asserting on exception messaging is not a very good idea in my opinion. Although @joshp123 recommended using a `RuntimeError`, it might be even better to create a custom exception class and use that.

lib/database_config.py
cwesdorp: Change to triple-quote. If people are checking build pack code I will assume they know what a VCAP is.

tests/usecase/test_db_config_options.py
```diff
@@ -0,0 +1,85 @@
...
+        assert config is None
+
+    def test_database_url(self):
+        self.clean_env()
+        os.environ[
+            "DATABASE_URL"
+        ] = "jdbc:postgres://user:secret@host/database"
+
+        config = get_database_config()
+        assert config is not None
```
matyaskuti: Could you extend this beyond making sure that the config is not `None`? If someone was to remove almost all logic from `get_database_config` and only return an empty string for example, this test would still happily pass, even though the `config` is not what we expect.

Also for the next testcase.

lib/database_config.py
cwesdorp: get_m2ee_configuration returns None, "if m2ee_config will work here.

tests/usecase/test_db_config_options.py
cwesdorp: without the list it throws "RuntimeError: dictionary changed size during iteration"

tests/usecase/test_db_config_options.py
cwesdorp: yes, but that is not the intension of this test. db specific logic is tested in other test cases.

tests/usecase/test_db_config_options.py
matyaskuti: I overlooked that, but makes sense because you are deleting items.

tests/usecase/test_db_config_options.py
matyaskuti: But don't we want to make sure that the configuration is assembled correctly based on environment variables? Just to be sure, could you link those tests?

tests/usecase/test_db_config_options.py
matyaskuti: 👍
One last thing here: since `config` is not used, you can remove the assignment and just leave the function call.

matyaskuti:
@cwesdorp For future ease of git blaming, could you split up the last commit with the review related changes into multiple ones explaining the what and why of the separate logical changes?

tests/usecase/test_db_config_options.py
cwesdorp: specific cases are test_rds, test_mysql, test_sap_hana. Update the test to check the database type in the returned config.

tests/usecase/test_db_config_options.py
cwesdorp: that's done as well

README.md
cwesdorp: changed, added example as well.

lib/metrics.py
cwesdorp: changed

tests/usecase/test_db_config_options.py
cwesdorp: changed to context manager

lib/metrics.py
```diff
@@ -327,12 +328,19 @@ def _get_db_conn(self):
             self.db = None
 
         if not self.db:
+            # get_database config may return None or empty
             db_config = database_config.get_database_config()
+            if not db_config or "DatabaseType" not in db_config:
+                raise Exception(
```
joshp123: As previous, it's generally better to raise a RuntimeError as opposed to just an exception.

Although, here, a ValueError is probably better.

start.py
```diff
@@ -667,11 +667,15 @@ def set_runtime_config(metadata, mxruntime_config, vcap_data, m2ee):
...
-        database_config.get_database_config(
-            development_mode=is_development_mode()
-        )
+
+    # db configuration might be None, database should then be set up with
+    # MXRUNTIME_Database... custom runtime settings.
+    runtime_db_config = database_config.get_database_config(
+        development_mode=is_development_mode()
     )
+    if runtime_db_config is not None:
```
joshp123: As before, `if not runtime_db_config` is probably better here

lib/database_config.py
```diff
@@ -38,7 +58,7 @@ def get_instance(self):
 
         # fallback to original configuration
         url = self.get_database_uri_from_vcap(self.vcap_services)
-        if url is None:
+        if url is None and "DATABASE_URL" in os.environ:
```
joshp123: Consider `if not url`

lib/metrics.py
matyaskuti: Or even a custom, domain exception.

tests/usecase/test_db_config_options.py
```diff
@@ -0,0 +1,85 @@
+import os
```
joshp123: I would run `black` on this file. We don't enforce formatting for tests, but some of this could do with a black run.

lib/metrics.py
joshp123: Even better.

```
class MissingDatabaseConfigException(ValueError):
    pass

if not 'foo' in 'bar':
    raise MissingDatabaseConfigException("your message here")
```

cwesdorp: valuable suggestion, but this then should apply to all usage of database config. I make a new branch for these improvements to prevent blowing up the scope here okay?

joshp123: Sure, but in the meantime, using ValueError/RuntimeError instead of exception is still a good idea. (and easy)

tests/usecase/test_db_config_options.py
cwesdorp: actually black is run on the test files. what exactly do you mean here?

joshp123: Eh, ignore this, a few things looked wrong to my eye, but actually were not. Never mind!

lib/metrics.py
cwesdorp: Done.

lib/database_config.py
cwesdorp: done

start.py
cwesdorp: changed

tests/usecase/test_db_config_options.py
```diff
@@ -0,0 +1,85 @@
...
+        based on DATABASE_URL or VCAP_SERVICES
+        """
+        self.clean_env()
+        os.environ["MXRUNTIME_DatabaseType"] = "PostgreSQL"
+        os.environ[
+            "MXRUNTIME_DatabaseJdbcUrl"
+        ] = "postgres://username:password@rdsbroker-testfree-nonprod-1-eu-west-1.asdbjasdg.eu-west-1.rds.amazonaws.com:5432/testdatabase"  # noqa E501
+
+        config = get_database_config()
+        assert not config
```
matyaskuti: This implies that the `config` value is falsy. Is this the intention? So it's okay if the config is `0` or `""`? If not, then please explicitly check for `None`

cwesdorp: Yes, for this test that is okay. we just want to make sure we get no config.

PR #271, opened 2019-07-31
MxBuild process check improvements

cwesdorp:
During running unit tests for check_db_config_options a test with MxBuild failed twice. To investigate and improve this branch is created. 

Run tests with debug mode enabled. 
Have JDK tests with source push wait for mxbuild, moved wait code to test base class.
Check process status for MxBuild after start.


cwesdorp:
Not merging, looking for other options.

PR #282, opened 2019-10-18
Dep 2150 datadog warning

svanderburg:


lib/datadog.py
```diff
@@ -305,10 +305,14 @@ def compile(install_path, cache_dir):
...
 
-def run():
+def run(runtime_version):
     if not is_enabled():
         return
 
+    if runtime_version < 7.14:
+        logger.error("Datadog integration requires Mendix 7.14 or newer")
+        raise Exception("Datadog integration requires Mendix 7.14 or newer")
+
```
cwesdorp: Let's check this in the compilation phase of the build pack, and not during run. I'm also not sure an exception is needed stopping the app from running or just a big warning in the log.

tests/usecase/test_datadog.py
```diff
@@ -25,6 +25,13 @@ def _test_datadog_running(self, mda_file):
         assert output is not None
         assert str(output).find("datadog") >= 0
 
+    def test_datadog_failure_mx6(self):
+        super().setUp()
+        self.setUpCF(
+            "sample-6.2.0.mda", env_vars={"DD_API_KEY": "NON-VALID-TEST-KEY"}
+        )
+        self.startApp(expect_failure=True)
```
cwesdorp: Can we have an offline test to validate the mendix version.

Makefile
```diff
@@ -2,5 +2,5 @@ test:
 	cd tests && bash run.sh
 
 lint:
-	flake8 --exclude .git,__pycache__,lib/certifi,lib/idna,lib/psycopg2,lib/urllib3,lib/chardet,lib/httplib2,lib/requests,lib/yaml
+	flake8 --exclude .git,bin,lib,__pycache__,lib/certifi,lib/idna,lib/psycopg2,lib/urllib3,lib/chardet,lib/httplib2,lib/requests,lib/yaml
```
cwesdorp: bin and lib should be included, our code is there in as well

lib/datadog.py
```diff
@@ -305,10 +305,13 @@ def compile(install_path, cache_dir):
     download(install_path, cache_dir)
 
 
-def run():
+def run(runtime_version):
     if not is_enabled():
         return
 
+    if runtime_version < 7.14:
+        logger.error("Datadog integration requires Mendix 7.14 or newer")
```
cwesdorp: lets make this a warning. Append something like' Datadog agent is not activated.'

cwesdorp:
Approved.

PR #234, opened 2019-03-12
UP-777 set user provided java options

diogomatsubara:
This change was reviewed here: https://github.com/mendix/cf-mendix-buildpack/pull/232 by josh and the tests are passing (after I fixed the formatting): https://travis-ci.org/mendix/cf-mendix-buildpack/builds/505183476

PR #260, opened 2019-06-26
Add custom java_opts instruction to readme

Infra-Red:

PR #301, opened 2020-03-13
Clarify Java options configuration

jeohist:
This setting is discussed in #232 / #234 but isn't very well documented.
This clarifies the expected format used to configure Java options.


jeohist:
Oh hey, this is done much better by #260

PR #306, opened 2020-03-30
Download Mendix runtimes from download.mendix.com

markbastiaans:
This PR closes internal issue DEP-2858 and makes Mendix runtimes come from download.mendix.com.


markbastiaans:
Closing PR due to another solution incoming

PR #307, opened 2020-04-06
Changes for SameSite cookies in runtimes < 8.11

markbastiaans:
This PR is for internal issue DEP-2872 and "fixes" SameSite compliancy for Mendix applications with any runtime < 8.9.

### DIFF
--- a/buildpack/nginx.py
+++ b/buildpack/nginx.py
@@ -41,7 +41,7 @@ def set_up_files(m2ee):
         lines = "".join(fh.readlines())
 
     samesite_cookie_workaround = get_path_config(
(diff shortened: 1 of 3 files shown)


lib/nginx.py
```diff
@@ -116,6 +118,12 @@ def get_path_config():
         if config.get("client-cert") or config.get("client_cert"):
             client_cert = "auth_request /client-cert-check-internal;"
 
+        # Temporary fix for SameSite enforcement (runtime does not set this cookie yet)
+        if runtime_version < 8.9:
```
KevinVlaanderen, line 122: The proper fix will not be in Mx8.9, but we do aim for 8.10. So this should probably be < 8.10.

PR #308, opened 2020-04-15
Add deleting of old cached dependencies

markbastiaans:
This PR closed internal issue DEP-2887 and adds deleting of old dependencies out of the build cache. This should free up staging space for apps that are upgraded during their lifetimes (e.g. runtime and other dependencies).

### DIFF
--- a/tests/requirements.txt
+++ b/tests/requirements.txt
@@ -1,6 +1,7 @@
 backoff==1.10.0
 certifi==2019.3.9
 chardet==3.0.4
+cryptography==2.8
 idna==2.8
 nose-timer==0.7.4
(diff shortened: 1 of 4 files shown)


tests/requirements.txt
```diff
@@ -1,6 +1,7 @@
 backoff==1.10.0
 certifi==2019.3.9
 chardet==3.0.4
+cryptography==2.8
```
cwesdorp, line 4: Why is this added?

markbastiaans: This is a test requirement and if you don't specify this, Python 3.4 will break.

PR #311, opened 2020-05-11
WIP: DVAS-1155

joshp123:
WIP: trying to fix negative JVM metrics.

### DIFF
--- a/lib/m2ee/munin.py
+++ b/lib/m2ee/munin.py
@@ -5,8 +5,10 @@
 # http://www.mendix.com/
 #
 
+import itertools
 import os
 import string
+import warnings
 
 from m2ee.log import logger
 from . import smaps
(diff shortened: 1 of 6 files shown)


markbastiaans:
@joshp123 - see https://github.com/mendix/cf-mendix-buildpack/pull/312#issuecomment-629144674 for my opinion on this approach

lib/m2ee/munin.py
```diff
@@ -220,21 +231,135 @@ def get_stats_from_runtime(client, config):
...
+            # correct pool is the "Metaspace" pool at index 1.
+            stats["memory"]["permanent"] = memorypools[1]["usage"]
             stats["memory"]["eden"] = memorypools[3]["usage"]
             stats["memory"]["survivor"] = memorypools[4]["usage"]
             stats["memory"]["tenured"] = memorypools[5]["usage"]
     elif java_version >= 8:
+        # This branch should only be reached with a Mendix runtime version
+        # <= 6.5 (i.e. before MemoryPools were added), and one using Java 8,
+        # which seems to include all Mendix 6 runtimes, and some (but not all)
+        # Mendix 5 runtimes.
```
joshp123, line 362: @pappachino - this is the remaining case that needs to be re-implemented. This is the case where the original fields were all completely wrong. This is only the case for Mendix 6 and Java 8, without memorypools (The unit tests should make this clear, I hope)

tests/unit/test_java_version_parsing.py
```diff
@@ -109,3 +108,14 @@ def test_guess_mendix_5_java_8_with_missing_java_version(self):
             m2ee_about, runtime_version, stats
         )
         self.assertEqual(8, guessed_version)
+
+    def test_guess_future_mendix_versions_doesnt_error(self):
```
joshp123, line 112: I'm not sure what this test case was supposed to do actually. I think I was aiming to trigger the floor operator in `guess_java_version`?

joshp123:
This MR has been decomposed into some others. Closing it for now.

PR #312, opened 2020-05-12
Apply Black linting to lib/m2ee too.

joshp123:
This brings it in line with the rest of the buildpack code. (see https://github.com/mendix/cf-mendix-buildpack/pull/198)

Black checks the AST (abstract syntax tree) to ensure that code is identical, so this change is safe, and passing unit tests will confirm this.

- [x] Check everything in Contributing.md

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -57,7 +57,7 @@ install_lint_requirements: install_piptools
 
 .PHONY: lint
 lint:
-	black --line-length=79 --check --diff $(PACKAGE_NAME) tests/*/
+	black --line-length=79 --check --diff $(PACKAGE_NAME) lib/m2ee/* tests/*/
 	pylint --disable=W,R,C $(PACKAGE_NAME) tests/*/
 
 .PHONY: install_test_requirements
@@ -96,4 +96,4 @@ test: test_unit test_integration
 
 .PHONY: format
 format:
-	black --line-length=79 $(PACKAGE_NAME) tests
+	black --line-length=79 $(PACKAGE_NAME) tests lib/m2ee/*
--- a/buildpack/start.py
+++ b/buildpack/start.py
@@ -27,7 +27,7 @@ from buildpack import (
 from buildpack.runtime_components import security
 from lib.m2ee import M2EE as m2ee_class
 
-BUILDPACK_VERSION = "4.5.4"
+BUILDPACK_VERSION = "4.5.5"
 
 m2ee = None
 app_is_restarting = False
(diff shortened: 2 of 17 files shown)


markbastiaans:
@joshp123 - I would suggest building your m2ee fixes on the upstream version of m2ee-tools. The version in the buildpack was patched by us in a murky past for Python 3 compatibility, which is in the most recent upstream version (I did a diff to confirm). IMO there is no sense in branching off further on the old m2ee code that is in the buildpack.

What we can then do is package the upstream version as a wheel in the build pipeline of the buildpack, and then include it as we include other Python dependencies.

joshp123:
I totally agree that the two divergent versions of m2ee-tools should converge again, and become one single repository. 

The reason that I chose to begin from the buildpack (v4) version of m2ee tools for the changes in #311, and not the separate (v3) repository, is that the two versions already seem to be quite divergent. For example if you compare the commit history of the two:
https://github.com/mendix/cf-mendix-buildpack/commits/master/lib/m2ee
https://github.com/mendix/m2ee-tools/commits/master
then a cursory looks seems to suggest that the last commit to integrate the two forks was this one, in 2016: https://github.com/mendix/cf-mendix-buildpack/commit/3cefd0277b56691413b17b5c74a342c9a5c1b4a8#diff-00e4c83efd57f0eb4eb6a4b215ba343e, and the (v3) m2ee-tools repository has had ~100 commits since that time; whilst the v4 buildpack has had about 10ish m2ee-tools commits.

Unfortunately that makes things a little more complicated for us, as re-integrating the two will require us checking (and testing) to make sure that changes in one don't introduce incompatibilities for the other. Based on my fairly limited knowledge, I think that there are a *lot* of incompatible changes, and fixing it will be a decent amount of work.

Therefore, if I were to have based my changes on the v3 version of m2ee-tools, it would necessitate that we re-integrate the v3 and v4 m2ee-tools version before being able to fix this issue for the customer.

With regards to the best way forward here, I am fine with pausing code-quality improvements like Black formatting, Pylint, etc, but I think that (for example), if we add unit tests here (as in #311) they will need to be ported across to the other m2ee tools repository.

My suggestion:
- Drop code quality changes, i.e. close this MR (even though it makes me sad)
- Apply our changes to v4 m2ee-tools in parallel (manually or by cherry-picking) to a separate "reintegration" branch to the v3 m2ee-tools repository, to at least keep the two somewhat in sync
- Raise this divergency during the next quarterly planning session, and put this reintegration on the roadmap
- Invest some roadmap time in analysing the scale of the reintegration (e.g. are our assumptions that this is very divergent correct)

What do you think?

pappachino:
@joshp123 I've added the lib/m2ee to the make `format` target too and bumped the version.

Could you have a look and approve if all good. 
Thanks.

joshp123:
LGTM! I don't think I can approve because I opened the MR and helped author one of the commits.

@markbastiaans - can you please take a look?

PR #313, opened 2020-05-18
DEP-2467: Configuration of Nginx. Configuring location access by cert…

carbone-kid:
…ificates issuers. Updated README.

### DIFF
--- a/DEVELOPING.md
+++ b/DEVELOPING.md
@@ -21,6 +21,16 @@ For developing the buildpack, `pyenv` must be set up, and the dependencies for t
 
 Beware that an older file system is also supported which includes Python 3.4, so please take that into account while introducing additional dependencies or new dependency version.
 
+### Developing in `Docker`
+As an alternative to running Python on your host you can run it in Docker container.
+* set up required environment variables
+* go to `./dev` folder
+* execute `bash run-locally.sh`
+
+This will start the Docker container with preinstalled Python and "ssh" you into it.
+The project folder will be mapped to the current folder in the Docker container, so if you edit files on you host 
+the changes will be immediately available in the container.
+
 ### Installing testing and linting requirements
 The buildpack makes use of the `make` system. For dependency management, `pip-compile` is used under the hood.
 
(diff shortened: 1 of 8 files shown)


DEVELOPING.md
```diff
@@ -21,6 +21,20 @@ For developing the buildpack, `pyenv` must be set up, and the dependencies for t
 
 Beware that an older file system is also supported which includes Python 3.4, so please take that into account while introducing additional dependencies or new dependency version.
 
+### Running in `Docker`
+As an alternative to running Python on your host you can run it in Docker container.
+* set up required environment variables
+* go to `./run-locally` folder
```
markbastiaans, line 27: I think we should check if the `run-locally` folder is the right place / name for the Dockerfile, composefile and bash script for the development environment. I would suggest putting these in a `dev` folder instead, but that's up for discussion.

DEVELOPING.md
```diff
@@ -21,6 +21,20 @@ For developing the buildpack, `pyenv` must be set up, and the dependencies for t
...
+* execute `bash run-locally.sh`
+
+This will start the Docker container with preinstalled Python and "ssh" you into it.
+The project folder will be mapped to the current folder in the Docker container, so if you edit files on you host 
+the changes will be immediately available in the container.
+
+To run a test separately do:
+
+`nosetests --nocapture --verbosity=3 --processes=10 --process-timeout=3600 --with-timer --timer-no-color tests/integration/test-file-name.py`
```
markbastiaans, line 36: This might be better put in the `Testing` section of this document.

buildpack/nginx.py
```diff
@@ -236,6 +238,18 @@ def get_path_config(samesite_cookie_workaround=False):
         if config.get("client-cert") or config.get("client_cert"):
             client_cert = "auth_request /client-cert-check-internal;"
 
+        issuer_dn = ""
```
markbastiaans, line 241: Will this break if we initialise this as `None`? Python is quite different than Java in handling "empty" values.

DEVELOPING.md
carbone-kid: Agree, dev is better

DEVELOPING.md
carbone-kid: Moved to Testing

buildpack/nginx.py
carbone-kid: I would leave this for now.

PR #318, opened 2020-05-25
SameSite cookie support for Chrome > 80

markbastiaans:
This PR supersedes the previous PR. It adds the SameSite and Secure attributes to all cookies for runtime versions < 8.12.

It explicitly does not set the HttpOnly attribute, which broke applications with custom handlers which used client-side cookie access.


jeohist:
@markbastiaans FYI: https://blog.chromium.org/2020/05/resuming-samesite-cookie-changes-in-july.html

README.md
```diff
@@ -206,6 +206,14 @@ to configure multiple supported headers, you can set it like below:
 cf set-env <YOUR_APP> HTTP_RESPONSE_HEADERS '{"Referrer-Policy": "no-referrer-when-downgrade", "X-Content-Type-Options": "nosniff"}'
 ```
 
+### Enabling SameSite / Secure Cookie Header Injection for Runtime < 8.11
+
+Google Chrome will - at a certain moment - [enforce cookie security](https://www.chromium.org/updates/same-site) by requiring the `SameSite` and `Secure` atrributes for all cookies. Mendix runtime versions < 8.11 do not include these properties in cookies.
+
+The buildpack can inject these two properties into all cookies for affected runtime versions.
+
+This workaround is disabled by default. If your application supports injecting these cookies, you can choose to enable cookie header injection by setting the `SAMESITE_COOKIE_WORKAROUND` environment variable to `true`.
```
KevinVlaanderen, line 215: Why would we actually have the runtime version check, if this fix is opt-in?

KevinVlaanderen: The workaround is needed for 8.11 as well, because of a client change we're doing for 8.12, but that could be part of the documentation.

buildpack/nginx.py
```diff
@@ -19,6 +21,35 @@
 }
 
 
+# Fix for Chrome SameSite enforcement (from Chrome 80 onwards)
+# Runtime will set this cookie in runtime versions >= SAMESITE_COOKIE_WORKAROUND_LESS_MX_VERSION
+SAMESITE_COOKIE_WORKAROUND_ENV_KEY = "SAMESITE_COOKIE_WORKAROUND"
```
KevinVlaanderen, line 26: Should we call it "workaround"? Since this is a customer-facing environment option, it may not have the right ring to it. Perhaps "SAMESITE_COOKIE_PRE_MX812".

markbastiaans: I agree, will change this

README.md
markbastiaans: We don't want people enabling this and when 8.12 arrives struggling with finding why their custom cookie settings in 8.12 are not working.

markbastiaans: I'll change the 8.11 to 8.12.

PR #319, opened 2020-05-29
DVAS-1236: Fix guess_java_version function for Java 11.

pappachino:
Add unit tests to cover all of the strange legacy edge cases. 
Refactor a little to make the function more unit testable.

### DIFF
--- a/lib/m2ee/munin.py
+++ b/lib/m2ee/munin.py
@@ -26,6 +26,7 @@ except ImportError:
         )
         raise
 
+
 default_stats = {
     "languages": ["en_US"],
     "entities": 0,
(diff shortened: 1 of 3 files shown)


lib/m2ee/munin.py
```diff
@@ -104,35 +105,63 @@ def print_config(m2ee, name):
...
+    """
+    if not m2ee_response.has_error():
+        about = m2ee_response.get_feedback()
+        if "java_version" in about:
+            java_version_string = about["java_version"]
+            return _get_jre_major_version_from_version_string(
+                java_version_string
+            )
+
+    # These branches should never be reached for Mendix versions in the Mendix
```
joshp123, line 158: I wrote this comment, but I think that it's not actually accurate. You can run all Mendix 6 versions in Cloud V4, some of which lack the "java_version" field.

I think that this comment should only refer to line 162,  `if runtime_version // 5`

Can you verify this with an early Mendix 6 version and share the m2ee about output? (The "188v4test" application in the test environment has a Mendix 6.5 MDS deployed to the "test" environment of that application, that might be a good starting point).

tests/unit/test_java_version_parsing.py
```diff
@@ -0,0 +1,111 @@
...
+        stats = {}
+        guessed_version = _guess_java_version(
+            m2ee_about, runtime_version, stats
+        )
+        self.assertEqual(11, guessed_version)
+
+    def test_guess_mendix_6_with_missing_java_version(self):
+        """All Mendix 6 versions are supposed to use Java 8.
+
+        This code is deprecated, but I am including unit tests against the
```
joshp123, line 60: Let's update this deprecation warning based on the investigation above.

lib/m2ee/munin.py
pappachino: @joshp123 This is how the about output looks like : 
```
{'feedback': {'build': '8913',
              'copyright': 'Copyright © 2003-2016 Mendix bv. All rights '
                           'reserved.',
              'model_version': 'unversioned',
              'name': 'Mendix Runtime',
              'vendor': 'Mendix',
              'version': '6.5.0',
              'xasid': 'b82e1c10-dd51-4a0d-8998-e1361d903ca1'},
 'result': 0}
```

Yeah, I don't see the `java_version` string, but we would reach that branch of course. I'll move the comment to the line below.
I'll move the line for runtime_version

pappachino: I've ported the relevant commit changes from #320 which has the appropriate comments.

tests/unit/test_java_version_parsing.py
pappachino: Will update

lib/m2ee/munin.py
```diff
@@ -125,20 +126,53 @@ def print_values(m2ee, name):
...
+    # java_version field was added somewhere between Mendix 6.5 (not present)
+    # and Mendix 6.10.10 (present). The exact version that it was added in
+    # is probably not important.
+    # Anyway, all Mendix 6 runtimes run on Java 8.
+    if runtime_version.major == 6:
         return 8
+
+    # This branch should never be reached for Mendix versions in the Mendix
+    # Cloud, see also https://github.com/mendix/m2ee-tools/issues/51
     if runtime_version // 5:
```
joshp123, line 168: Can we also use `runtime_version.major` here, like above? I think that that is safer :) 

```suggestion
    if runtime_version.major == 5:
```

lib/m2ee/munin.py
```diff
@@ -125,20 +126,53 @@ def print_values(m2ee, name):
...
-        m = stats["memory"]
+        m = m2ee_stats["memory"]
         if m["used_nonheap"] - m["code"] - m["permanent"] == 0:
             return 7
         return 8
+
+    # If we have reached here, just return none as we were not
+    # able to guess the java version based on the available data.
```
joshp123, line 175: I think that this preserves the existing behaviour, so it should be safe. @markbastiaans - it would be good to have a second opinion here, though.

joshp123:
@markbastiaans - would you mind taking a look here too, please? Thank you in advance!

lib/m2ee/munin.py
markbastiaans: LGTM

pappachino:
_Release Notes:_

This release contains a bug fix in the `m2ee` library.

Bug fix addresses the java version detection logic mainly to account for the JDK version-string scheme changes introduced as per [JEP-223](https://openjdk.java.net/jeps/223)

PR #321, opened 2020-06-04
DVAS-1240: Raise RuntimeError for negative metrics

pappachino:
Having negative metrics displayed does not make sense. If there are any more bugs in our metrics, Instead of storing negative metrics, we should throw an error, and request that the customer contacts support.

This will show blank metrics, instead of erroneous metrics. There will also be a clear message in the customer’s application logs, explaining that something is wrong. That will give the customer more certainty than the current situation, where they cannot trust their graphs.

### DIFF
--- a/tests/unit/test_emit_metrics.py
+++ b/tests/unit/test_emit_metrics.py
@@ -2,7 +2,66 @@ import copy
 from unittest import TestCase
 from unittest.mock import Mock
 
-from buildpack.runtime_components.metrics import FreeAppsMetricsEmitterThread
+from buildpack.runtime_components.metrics import (
+    FreeAppsMetricsEmitterThread,
+    PaidAppsMetricsEmitterThread,
+)
+
+
+class TestNegativeMemoryMetricsThrowError(TestCase):
+    def test_validating_bad_metrics(self):
+        m2ee_stats = {"memory": {"javaheap": -12345}}
+        with self.assertRaises(RuntimeError):
+            PaidAppsMetricsEmitterThread._sanity_check_m2ee_stats(m2ee_stats)
+
+    def test_no_memorypools_good_metrics(self):
+        m2ee_stats = {"memory": {"javaheap": 12345}}
+        self.assertIsNone(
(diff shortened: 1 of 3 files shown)


joshp123:
I think that we can probably rebase this into a single commit, but other than that, looks good.

tests/unit/test_emit_metrics.py
```diff
@@ -2,7 +2,60 @@
...
+    FreeAppsMetricsEmitterThread,
+    PaidAppsMetricsEmitterThread,
+)
+
+
+class TestNegativeMemoryMetricsThrowError(TestCase):
+    def test_validating_bad_metrics(self):
+        m2ee_stats = {"memory": {"javaheap": -12345}}
+        with self.assertRaises(RuntimeError):
+            PaidAppsMetricsEmitterThread._sanity_check_m2ee_stats(m2ee_stats)
```
joshp123, line 15: I think that we are technically missing one test case here, the happy path where `m2ee_stats` only contains a positive int, and nothing else, right?


(The `test_non_ints_dont_cause_problems` function kinda does the same thing but not really; that is more checking that if you have a dict in `m2ee` tools as well as int values, no errors are raised)

pappachino: Good point. I have added a test for no memory pools, but good metrics.

pappachino:
Release Notes : Raise an error if we encounter negative memory metrics

If there are any more bugs in our metrics, instead of storing negative metrics, this release would throw an error, and request that the customer contacts support.
This would allow the developers to catch any other metrics related bugs early.

joshp123:
@pappachino suggest that we tweak the release notes slightly:

> Release Notes : Raise an error if we encounter negative memory metrics.
> Having negative metrics displayed does not make sense. If there are any more bugs in our metrics, Instead of storing negative metrics, we should throw an error, and request that the customer contacts support.
> 
> This will show blank metrics, instead of erroneous metrics. There will also be a clear message in the customer’s application logs, explaining that something is wrong. That will give the customer more certainty than the current situation, where they cannot trust their graphs.

PR #322, opened 2020-06-09
DVAS-1246 postgresql input plugin for telegraf

umangdhaw:
* Add input plugin in telegraf config for postgresql
* If **APPMETRICS_TARGET** is set, the database metrics will be enabled by default in telegraf. To opt out, set **APPMETRICS_INCLUDE_DB** to **False**

Tested by configuring an output plugin to write to file and verifying that metrics are being written.


umangdhaw:
@cwesdorp Request you to please review and merge

PR #325, opened 2020-06-18
DVAS-1237 : Refactor metric parsing

pappachino:
- Refactor the current "get_stats_from_runtime" function to extract metrics from memorypools by names, not by order.
- Account for variations in the pool names, based on the garbage collector enabled
- Add unit tests to assert that the new and old implementations produce the same results where appropriate

### DIFF
--- a/buildpack/start.py
+++ b/buildpack/start.py
@@ -27,7 +27,7 @@ from buildpack import (
 from buildpack.runtime_components import security
 from lib.m2ee import M2EE as m2ee_class
 
-BUILDPACK_VERSION = "4.5.8"
+BUILDPACK_VERSION = "4.6.1"
 
 m2ee = None
 app_is_restarting = False
--- a/lib/m2ee/munin.py
+++ b/lib/m2ee/munin.py
@@ -5,8 +5,10 @@
 # http://www.mendix.com/
 #
 
+import itertools
 import os
 import string
+import warnings
 
 from m2ee.log import logger
 from . import smaps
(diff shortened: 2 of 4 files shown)


buildpack/start.py
```diff
@@ -27,7 +27,7 @@
 from buildpack.runtime_components import security
 from lib.m2ee import M2EE as m2ee_class
 
-BUILDPACK_VERSION = "4.5.7"
+BUILDPACK_VERSION = "4.6.0"
```
joshp123, line 30: I think that because the 4.6.0 tag was already created in the previous release, we need to go to 4.6.1 here.

lib/m2ee/munin.py
```diff
@@ -229,21 +231,185 @@ def get_stats_from_runtime(client, config):
...
+        "Survivor Space": (
+            "PS Survivor Space",
+            "Par Survivor Space",
+            "G1 Survivor Space",
+        ),
+        "Tenured Gen": ("PS Old Gen", "CMS Old Gen", "G1 Old Gen",),
+    }
+
+    for standard_name, valid_names in pool_name_aliases.items():
+        for name in valid_names:
```
joshp123, line 371: I think that this would be slightly easier to read (at least for me) if we reversed the test, and also renamed `pool_name` to `given_pool_name`, i.e.:

```
for name in valid_names:
    if name == given_pool_name:
        return standard_name
```

What do you think?

pappachino: Sure. I'll update it

buildpack/start.py
markbastiaans: Indeed we need to go to v4.6.1, re-using tags in git is messy.
Also, we should re-consider adding the version number in code, maybe generating from a git ref / tag during the build process is the way to go here.

pappachino: I've bumped the version to 4.6.1

PR #327, opened 2020-06-22
DVAS-1241: Bug fix to address negative metrics issue

pappachino:
- Re-arrange the code a bit, to have minimal control flow branching

- Code cache is allocated out of native memory and
is not part of the java heap. So as previously thought
we should not remove the code cache space from java heap
when calculating the unused java heap.

- Similarly, since Java 8 metaspace (or permanent gen)
is allocated out of the native memory. So we should not
negate that space from the native memory calculations.

### DIFF
--- a/buildpack/start.py
+++ b/buildpack/start.py
@@ -27,7 +27,7 @@ from buildpack import (
 from buildpack.runtime_components import security
 from lib.m2ee import M2EE as m2ee_class
 
-BUILDPACK_VERSION = "4.6.1"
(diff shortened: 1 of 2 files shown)


pappachino:
Release Notes: Bug fix to address negative metrics issue

In this release we have couple of minor adjustments in the way we calculate _unused java heap_
and _native memory_
- exclude code cache from heap calculation.
- exclude metaspace (or permanent gen) from native memory calculation.

PR #329, opened 2020-06-26
Use MemoryMXBean's committed_heap metric to calculate free unused java heap

pappachino:
Currently we calculate the free _unused java heap_ metric by comparing `java heap` value from the `/proc/PID/smaps` file and the `used_heap` value from the `MemoryMxBean` interface. This at times results in `negative unused java heap`

Instead of relying on two different methodologies to gather metrics, simply use the `committed_heap` metric we have from the
[MemoryMXBean interface](https://cr.openjdk.java.net/~iris/se/11/latestSpec/api/java.management/java/lang/management/MemoryUsage.html) to calculate the free unused java heap.

Add tests to ensure we calculate free unused heap using committed_heap.

Bump version 4.7.0 → 4.7.1

### DIFF
--- a/tests/unit/test_metric_parsing.py
+++ b/tests/unit/test_metric_parsing.py
@@ -1,13 +1,17 @@
 import copy
 from unittest import TestCase
+from unittest.mock import Mock
 
+from lib.m2ee import smaps
 from lib.m2ee.munin import (
     _populate_stats_by_java_version,
     _populate_stats_by_java_version_old,
     _standardize_memory_pools_output,
+    augment_and_fix_stats,
 )
 
-
+# Memory stats from the runtime
+# → client.runtime_statistics().get_feedback()["memory"]
 MENDIX_5_6_0_JAVA_7 = {
     "code": 2034048,
     "committed_heap": 518979584,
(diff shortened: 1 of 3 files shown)


tests/unit/test_metric_parsing.py
```diff
@@ -186,6 +190,7 @@
 }
 
 MENDIX_8_JAVA_11_STATS_PARALLELGC = {
+    "code": 0,
```
joshp123, line 193: What is the reason for adding "code" here to this dict? Originally this dict was something that I copied from the output of the calls direct to the runtime, so I'm curious why we are changing it :)

tests/unit/test_metric_parsing.py
```diff
@@ -783,3 +788,102 @@ def test_poolnames_g1gc(self):
...
+            2: 193744,
+            3: 3216,
+            4: 0,
+            5: 92176,
+            6: 0,
+        }
+
+        smaps.get_smaps_rss_by_category = Mock(return_value=mock_smaps_output)
+
+        # Check for following Java 1 scenarios:
```
joshp123, line 862: Java 1? 😉

tests/unit/test_metric_parsing.py
pappachino: Oh ok. I thought we had missed the 'code' keys in Java 11, as I see it in the older Java version dicts.
But I just confirmed, it indeed is missing from the runtime stats.

I'll remove those.

pappachino:
Release Notes: Improvement to metrics in Mendix Cloud V4

We have changed the way that the unused java heap metric is calculated to make the data more accurate.
This also fixes the issue where this metric could have a negative value.

This value is displayed in the following graphs in Metrics for Mendix Cloud v4:
- as unused in [JVM Object Heap](https://docs.mendix.com/developerportal/operate/trends-v4#a-name-trends-appmxruntimejvmheap-a-4-4-jvm-object-heap)
- as unused java heap in [JVM Process Memory Usage](https://docs.mendix.com/developerportal/operate/trends-v4#a-name-trends-appmxruntimejvmprocessmemory-a-4-5-jvm-process-memory-usage)

This change will take effect the next time you deploy your application.

NOTE : You may see a large increase in the size of the unused java heap metric when this change is applied. This is a consequence of the new calculation and not a change in the memory usage of your app

PR #336, opened 2020-08-03
Fix smaps parser to be compatible with the latest java version

alinagol:
Java version was bumped in the newest release of cf-mendix-buildpack, and that changed the format of `/proc/<pid>/smaps` file. Blocks that were previously labeled as `heap` moved to the beginning of the file, and the parser could not process not only these blocks but also all the next ones, as the first ones were marked with `stage=STAGE_BORK`.

The MR proposes a fix to this problem. With that change, the labels stay the same as they were before the upgrade. 

**Testing**
Tested with different smaps files from apps with both old and new buildpacks.

**_Option 1_**
- Choose an app and 
```console
cf ssh <env_id>
```
- Find PID 
```console
export PID=`ps aux | grep java | grep -v grep | awk '{print $2}'`
```
- Export smaps to a file 
```console
cat /proc/$PID/smaps > smaps.file
```
- Copy that file to your local env 
```console
scp -P 2222 -o User=cf:<app_guid>/0 ssh.bbl-<cluster>.<region>.mendixcloud.net: smaps.file .
```
- Parse the file 
```console
python lib/m2ee/smaps.py < smaps.file
```
- Make sure the values are not all 0s

**_Option 2_**
Alternatively, you can deploy an app [using the version of cf-mendix-buildpack from this MR](https://gitlab.rnd.mendix.com/cloud-unit/PaaS/wikis/How-to-test-CF-Mendix-Buildpack-changes#deploy-app-via-cli-), and run:
```console
export PID=`ps aux | grep java | grep -v grep | awk '{print $2}'`
python3 ./app/lib/m2ee/smaps.py < /proc/$PID/smaps
```


pappachino:
Here are couple of test results before and after applying the parser changes from the PR

# App under Mendix Runtime 7.3.0

Before the changes
---
Native code: 0 kB
Native heap and memory arenas: 0 kB
JVM Heap: 0 kB
Thread stacks: 0 kB
JAR files: 0 kB
Other: 0 kB

After the changes
---
Native code: 15324 kB
Native heap and memory arenas: 57804 kB
JVM Heap: 161604 kB
Thread stacks: 3280 kB
JAR files: 12024 kB
Other: 50316 kB

# App under Mendix runtime 8.7.0

Before the changes
---
Native code: 0 kB
Native heap and memory arenas: 0 kB
JVM Heap: 0 kB
Thread stacks: 0 kB
JAR files: 0 kB
Other: 0 kB

After the changes
---
Native code: 20004 kB
Native heap and memory arenas: 76368 kB
JVM Heap: 188880 kB
Thread stacks: 3060 kB
JAR files: 0 kB
Other: 87060 kB

The parser changes indeed fixes the bug introduced by the last buildpack changes.

PR #337, opened 2020-08-10
Fix for custom 403 error pages for PBAR on root

markbastiaans:
This PR fixes the accessibility of 403 error pages for cases where a path-based access restriction was set for the root path.


svanderburg:
@markbastiaans is it also possible to introduce a testcase that reproduces the issue we're trying to fix?

PR #341, opened 2020-09-03
Dep 3051 rollback rds changes

svanderburg:


mbyregowMx (approved):
rolling back rds changes

PR #350, opened 2020-10-06
Deploy Built Version on GitHub Releases

markbastiaans:
The Mendix Cloud Foundry Buildpack is now available on Github Releases as a "built version".

This ensures that you always have a prebuilt and minimized buildpack (ZIP file) to use for your deployments. We ensure that this is a runnable version of the buildpack with all required bootstrap dependencies included.

To use the release artifacts, please use the following link and replace `<RELEASE>` with the release you want to use:

```
https://github.com/mendix/cf-mendix-buildpack/releases/download/<RELEASE>/cf-mendix-buildpack.zip
```

**If you're version pinning, be sure to point to the GitHub Releases artifact from now on if it is available for the release you want to use. We do not guarantee that pinning to the repository itself will produce a usable buildpack in the future.**


README.md
```diff
@@ -38,16 +37,18 @@ Install the Cloud Foundry command line executable. You can find this on the [rel
 
 ### Push your app
 
-We push an mda (Mendix Deployment Archive) that was built by the Mendix Business Modeler to Cloud Foundry.
+We push an MDA (Mendix Deployment Archive) that was built by the Mendix Business Modeler to Cloud Foundry.
```
svanderburg, line 40: - Business Modeler is called Studio Pro nowadays

README.md
```diff
@@ -38,16 +37,18 @@ Install the Cloud Foundry command line executable. You can find this on the [rel
...
+cf push <YOUR_APP> -b <LINK-TO-BUILDPACK> -p <YOUR_MDA>.mda -t 180
 ```
 
-We can also push a project directory. This will move the build process (using mxbuild) to Cloud Foundry:
+We can also push a project directory. This will move the build process (using MxBuild) to Cloud Foundry:
 
 ``` shell
-cd <PROJECT DIR>; cf push -b https://github.com/mendix/cf-mendix-buildpack -t 180
+cd <PROJECT DIR>; cf push -b <LINK-TO-BUILDPACK> -t 180
 ```
```
svanderburg, line 52: Explanation is clear to me, but a bit abstract. In my own experience with open source projects, users may get confused if we don't give any concrete examples. This saves us some headaches.

markbastiaans: There's an explanation right at the top explaining how to substitute that in a command.
Also, I've added a disclaimer that users should at least be familiar with the Cloud Foundry CLI.

svanderburg (approved):

PR #352, opened 2020-10-13
Experimental Databroker Integration

markbastiaans:
This change adds the new experimental Databroker module (part of the Mendix DataHub platform).

The module adds a Kafka connect sidecar along with Debezium to support Materialized Virtual Entities (MVEs). This includes:
- A Kafka streams sidecar for any produce side mappings and transformations
- An update to the Datadog sidecar to include support for the Databroker components

**The Databroker module is currently experimental and is disabled by default.**

_Main contributor:_ @abhishekamte-mx


pommi:
Can the cruft in f8b971dea5 be removed using git rebase maybe? The only line that matters is `dynatrace,` on line https://github.com/mendix/cf-mendix-buildpack/pull/352/commits/f8b971dea52bf5259c6d1c86a84673cb59ca1779#diff-a14739c91f06dff6c3790b275157dc644c7ae94ead3b6bd20056119948dc16ccR11

buildpack/databroker/__init__.py
```diff
@@ -0,0 +1,153 @@
...
+
+def should_run_kafka_connect():
+    try:
+        if (
+            os.environ["CF_INSTANCE_INDEX"] != None
+            and int(os.environ["CF_INSTANCE_INDEX"]) == 0
+        ):
+            return True
+        else:
+            return False
```
svanderburg, line 60:
```python
return (os.environ["CF_INSTANCE_INDEX"] != None and int(os.environ["CF_INSTANCE_INDEX"]) == 0)
```
Nitpicking. It can be expressed more concisely, if desired. I'm typically not a big fan of explicitly returning `True` (and `False` in the else) statement.

svanderburg: My suggestion turns it into a one liner ;)

markbastiaans: @abhishekamte-mx - can you respond?

abhishekamte-mx: sure i can make the change, its a good suggestion to make the code more "pythonic". any other comments @svanderburg  @markbastiaans ?

svanderburg (approved):

buildpack/databroker/__init__.py
abhishekamte-mx: i have incorporated the change

PR #364, opened 2020-11-12
Add Action for Veracode integration

Anvay-Rajhansa:
This change adds a Github Action which will get executed on every merge to master and will push the buildpack code to Veracode for security scanning.

This change is part of internal ticket DEP-3201.

### DIFF
--- a/.github/workflows/veracodescan.yml
+++ b/.github/workflows/veracodescan.yml
@@ -1,0 +1,41 @@
+name: Push package for veracode scan
+
+on: 
+  pull_request:
+    types: [closed]
+    branches:
+      - master
+
+jobs:
+  pre:
+    runs-on: ubuntu-latest
+    outputs:
+      skip: ${{ steps.skip-check.outputs.should_skip }}
+    steps:
+    - id: skip-check
+      uses: fkirc/skip-duplicate-actions@master
+      with:
+        github_token: ${{ github.token }}
+        paths_ignore: '["**.md", "dev/**"]'
+  veracode-scan:
+    name: veracode scan
+    runs-on: ubuntu-latest
+    needs: pre
+    steps:
+      - name: Download pull request artifact
+        uses: dawidd6/action-download-artifact@v2
+        with:
+          workflow: test.yml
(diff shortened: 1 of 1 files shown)


markbastiaans:
A few comments:

- You're now zipping all the source code and sending it to Veracode, while there is already an artifact built in the pull request. The Deploy to Github Releases actions contains logic to fetch that artifact - let's replace the manual zipping with what actually is released.
- Have you seen the Github Actions Veracode integration [here](https://github.com/marketplace/actions/veracode-upload-and-scan)? If we can use that, this will save us on maintaining the custom Java actions
- Is there a Veracode badge available for the README? That would be a really nice way of linking the repo and Veracode results together.
- Please read the contribution guidelines on e.g. squashing your commits - please do so here.

.github/workflows/veracodescan.yml
```diff
@@ -0,0 +1,41 @@
...
+      - name: Download pull request artifact
+        uses: dawidd6/action-download-artifact@v2
+        with:
+          workflow: test.yml
+          pr: ${{github.event.pull_request.number}}
+          name: dist
+      - name: Upload and scan
+        uses: veracode/veracode-uploadandscan-action@master
+        with:
+          filepath: 'cf-mendix-buildpack.zip'
```
markbastiaans, line 34: Will this work? The path is (correct me if I'm wrong) `dist/cf-mendix-buildpack.zip`

markbastiaans (changes requested):
I've added one comment.
Also, be sure to test the action run. Thanks!

.github/workflows/veracodescan.yml
SwapnilThange: @markbastiaans I have tested with the above path _{filepath: 'cf-mendix-buildpack.zip'}_ and it was able to upload the package successfully. 
Earlier when I was using the **run** command to do the veracode scan, i tried using dist/cf-mendix-buildpack.zip but it gave me an error saying package not found hence I changed it to above one

markbastiaans (approved):

PR #369, opened 2020-11-18
Databroker: added support for producer side mappings

abhishekamte-mx:
- updated config generator to support metadata.json for producer side mappings
- added id to column whitelist for debezium
- added support for different logging levels for databroker module
- moved databroker components' monitoring into their own  datadog instances
- Updated streams sidecar to v0.23.0-9
- added an option to override the streams sidecar version
- updated itest to use custom mda


buildpack/databroker/config_generator/templates/jmx/kafka_connect.yaml
```diff
@@ -0,0 +1,208 @@
...
+        QueueTotalCapacity:
+          alias: debezium.postgres.streaming.queue-total-capacity
+        SourceEventPosition:
+          alias: debezium.postgres.streaming.source-event-position
+        TotalNumberOfEventsSeen:
+          alias: debezium.postgres.streaming.total-number-of-events-seen
+      bean_regex: 'debezium\.postgres:type=connector-metrics,context=streaming,server=.*'
+  host: localhost
+  port: 11003
+  java_bin_path: /home/vcap/app/.local/bin/java
```
markbastiaans, line 206: This is a hardcoded path, which will cause problems in the future. Please consider making this a template value.

buildpack/databroker/config_generator/templates/jmx/kafka_streams.yaml
```diff
@@ -0,0 +1,109 @@
...
+          suppression-emit-total:
+            alias: kafka.streams.processor.suppression-emit-total
+        tags:
+            thread-id: $1
+            task: $2
+            processor-node-id: $3
+        bean_regex: 'kafka\.streams:type=stream-processor-node-metrics,thread-id=(.*?),task-id=(.*?),processor-node-id=(.*?)(?:,|$)'
+  host: localhost
+  port: 11004
+  java_bin_path: /home/vcap/app/.local/bin/java
```
markbastiaans, line 107: This is a hardcoded path, which will cause problems in the future. Please consider making this a template value.

tests/integration/test_databroker.py
```diff
@@ -80,10 +81,14 @@ def test_databroker_running(self):
         )
         self._start_kafka_cluster()
 
+        os.environ[
+            "PACKAGE_URL"
+        ] = "https://s3-eu-west-1.amazonaws.com/databroker.dev/test-app/MyProducer902.mda"
```
markbastiaans, line 86: Have you thought about the implications of making your development blobstore URL public, and not fronting it with CloudFront?

Please find an alternative.

buildpack/util.py
```diff
@@ -158,6 +158,10 @@ def get_buildpack_loglevel():
     return logging.INFO
 
 
+def is_buildpack_debug_logging_enabled():
```
markbastiaans, line 161: Is this utility function really necessary?
I see two usages which can be replaced with either a logic evaluation or something like `str()`.

markbastiaans (commented):
I've added a few comments. Please address these and I will review again.

buildpack/databroker/config_generator/templates/jmx/kafka_connect.yaml
abhishekamte-mx: this has been changed to a template value

tests/integration/test_databroker.py
abhishekamte-mx: this has been changed to a template value

buildpack/util.py
abhishekamte-mx: this has been removed

markbastiaans (approved):

PR #373, opened 2020-11-24
Reorganize dependencies

markbastiaans:
This change reorganises dependencies. The following changes have been done:

- Moved over all unmanaged dependencies to managed dependencies from the Mendix CDN
- Extracted both the Datadog Java Agent and Mendix Java Agent as separate dependencies
- Moved file copy logic from `stage.py` to individual buildpack components
- Vendorized the Mendix Log Filter dependency


buildpack/runtime_components/logs.py
```diff
@@ -36,7 +37,15 @@ def run(self):
             while True:
                 proc = subprocess.Popen(
                     [
-                        "./bin/mendix-logfilter",
+                        str(
+                            os.path.abspath(
+                                os.path.join(
+                                    ".local",
+                                    "mendix-logfilter",
```
ajouwens-mendix, line 44: Is this correct, using `mendix-logfilter` twice?

markbastiaans: Yes, the binary is inside the `mendix-logfilter` directory. The integration test also calls this binary and it's working.

ajouwens-mendix (approved):

PR #379, opened 2020-11-30
Adjust certificate DN logging

carbone-kid:
This change allows to compare the client certificate Issuer DN with the Issuer DN configured for the endpoint in ARP.


etc/nginx/conf/nginx.conf.j2
```diff
@@ -118,6 +123,7 @@ http {
             {% if location.issuer_dn_regex %}
             if ($http_ssl_client_i_dn ~* ^(?!({{ location.issuer_dn_regex }})$)(\w*)) {
                 # Enable logging to show ssl_client_i_dn
+                set $issuer_dn '{{ location.issuer_dn }}';
```
markbastiaans, line 126: How will this work with multiple locations? The log format is set on a root level.

markbastiaans (changes requested):
I'm having doubts about whether this will work for multiple locations.

Also, please be a bit more verbose in your initial PR comment - a short description of the change should suffice.

I've also based this PR against `develop` - be sure to get that in there.

etc/nginx/conf/nginx.conf.j2
carbone-kid: It should work with multiple locations because the scope of the variable is the HTTP request. From the Nginx documantation: `Requests do not interfere with each other even if they are referencing a variable with the same name.`

markbastiaans (approved):

PR #386, opened 2020-12-15
Add support for base64-encoded input of Certificate Authorities

MXClyde:
This PR adds the ability to use environment variable CERTIFICATE_AUTHORITIES_BASE64 to enter a base64-encoded version of CERTIFICATE_AUTHORITIES. This eases use of the buildpack from a CI/CD tool that only allows one-lined environment variables.


buildpack/runtime_components/security.py
```diff
@@ -63,6 +63,10 @@ def create_admin_user(m2ee, is_development_mode):
 
 def get_certificate_authorities():
     config = {}
+    cas64 = os.getenv("CERTIFICATE_AUTHORITIES_BASE64", None)
+    if cas64:
+        certificate_authorities = base64.b64decode(cas64)
```
markbastiaans, line 68: Have you considered trying to base64 decode the original environment variable first? If that passes, you can use that value - if not, you can use the original parsing behaviour.

markbastiaans (changes requested):
Some general comments:
- For this and future contributions, please base your PR on `develop`.
- Please ensure that the tests pass.
- Please squash your commits to keep the commit history tidy.

There's also a remark about the code too that should remove the necessity for an extra environment variable.

PR #388, opened 2020-12-24
Escape single quote in issuer DN

carbone-kid:
This hotfix fixes an error in certificate Path-Based Access Restrictions (PBAR) when the issuer DN contains single-quote (`'`) characters.


buildpack/nginx.py
```diff
@@ -290,6 +290,7 @@ def get_access_restriction_locations():
                 # Escape special characters
                 issuer = issuer.replace(" ", "\\040")
                 issuer = issuer.replace(".", "\\.")
+                issuer = issuer.replace("'", "\\'")
```
markbastiaans, line 293: This will actually become `\'` instead of `'` - is this the intention?

markbastiaans (changes requested):
One code comment / question.

Since this should warrant a hotfix, be sure to:
- Base on `master`
- Add a description in the PR
- Bump the version
- Choose the "Squash and merge" option when merging

After merging, there will be a draft release in the Releases section. Publishing it will release this version into the wild.

markbastiaans:
I've based this branch on `master` - we should not forget to rebase `develop` after releasing.

buildpack/nginx.py
carbone-kid: I beleave it should render `\'` in the nginx.conf and the check should work for `'`

markbastiaans: Okay. Maybe it's good to create an integration test which verifies this. It's quite easy to spin up an application, apply a path restriction and do a HTTP request with an Issuer DN header set.

markbastiaans (approved):

PR #419, opened 2021-04-20
Support for fetching STS token from TVM for mendix runtime version >= 9.2

manishjindal:
This PR includes the following changes:
- Support for using STS token for accessing S3 bucket for Mendix Runtime >= 9.2

### DIFF
--- a/buildpack/runtime_components/backup.py
+++ b/buildpack/runtime_components/backup.py
@@ -25,8 +25,6 @@ def run():
     if amazon_s3 in vcap_services:
         s3_credentials = vcap_services[amazon_s3][0]["credentials"]
         backup_service["filesCredentials"] = {
-            "accessKey": s3_credentials["access_key_id"],
-            "secretKey": s3_credentials["secret_access_key"],
             "bucketName": s3_credentials["bucket"],
         }
         if "key_suffix" in s3_credentials:  # Not all s3 plans have this field
--- a/buildpack/runtime_components/storage.py
+++ b/buildpack/runtime_components/storage.py
@@ -2,15 +2,15 @@ import json
 import logging
 import re
 import os
-
+import requests
+import time
 from buildpack import util
 
 
 def _get_s3_specific_config(vcap_services, m2ee):
     access_key = secret = bucket = encryption_keys = key_suffix = None
-    endpoint = None
+    tvm_endpoint = tvm_username = tvm_password = endpoint = amazon_s3 = None
     v2_auth = ""
-    amazon_s3 = None
 
     blobstore_type = os.getenv("MENDIX_BLOBSTORE_TYPE")
 
@@ -23,9 +23,17 @@ def _get_s3_specific_config(vcap_services, m2ee):
 
     if amazon_s3:
         _conf = vcap_services[amazon_s3][0]["credentials"]
-        access_key = _conf["access_key_id"]
-        secret = _conf["secret_access_key"]
         bucket = _conf["bucket"]  # see below at hacky for actual conf
+        if "access_key_id" in _conf:
+            access_key = _conf["access_key_id"]
+        if "secret_access_key" in _conf:
+            secret = _conf["secret_access_key"]
+        if "tvm_endpoint" in _conf:
+            tvm_endpoint = _conf["tvm_endpoint"]
+        if "tvm_username" in _conf:
+            tvm_username = _conf["tvm_username"]
+        if "tvm_password" in _conf:
+            tvm_password = _conf["tvm_password"]
         if "encryption_keys" in _conf:
             encryption_keys = _conf["encryption_keys"]
         if "key_suffix" in _conf:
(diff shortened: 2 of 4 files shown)


buildpack/runtime_components/storage.py
```diff
@@ -65,17 +76,55 @@ def _get_s3_specific_config(vcap_services, m2ee):
...
+        config = {
+            "com.mendix.core.StorageService": "com.mendix.storage.s3",
+            "com.mendix.storage.s3.AccessKeyId": access_key,
+            "com.mendix.storage.s3.SecretAccessKey": secret,
+            "com.mendix.storage.s3.BucketName": bucket,
+        }
+    elif (tvm_endpoint and tvm_username and tvm_password and bucket):
+        if m2ee.config.get_runtime_version() < 9.2:
+            logging.info("S3 config detected, fetching IAM credentials from TVM")
+            response = requests.get(
```
pommi, line 90: I would use what is described here: https://docs.python-requests.org/en/master/user/authentication/
```
requests.get('https://api.github.com/user', auth=('user', 'pass'))
```
Then `_get_credential()` is not required.

buildpack/runtime_components/storage.py
```diff
@@ -65,17 +76,55 @@ def _get_s3_specific_config(vcap_services, m2ee):
...
+                "https://%s/v1/getcredentials" % tvm_endpoint,
+                headers={
+                "User-Agent": "Mendix Buildpack",
```
pommi, line 93: Could we include the buildpack version in the User-Agent string as well?

buildpack/runtime_components/storage.py
```diff
@@ -65,17 +76,55 @@ def _get_s3_specific_config(vcap_services, m2ee):
...
+                    "com.mendix.storage.s3.BucketName": bucket,
+                }
+            else:
+                logging.error("Failed to get AccessKeyID or SecretAccessKey from TVM")
+                return None
+        else:
+            logging.info("S3 config detected, activating external file store")
+            config = {
+                "com.mendix.core.StorageService": "com.mendix.storage.s3",
+                "com.mendix.storage.s3.tokenService.Url" : "https://%s/v1/gettoken" % tvm_endpoint,
```
pommi, line 121: Should this be `/v2/`?

buildpack/runtime_components/storage.py
```diff
@@ -65,17 +76,55 @@ def _get_s3_specific_config(vcap_services, m2ee):
...
+                "com.mendix.storage.s3.tokenService.Username" : tvm_username,
+                "com.mendix.storage.s3.tokenService.Password" : tvm_password,
```
pommi, line 123: Additionally I would advise to configure:
```
"com.mendix.storage.s3.tokenService.RefreshPercentage": 80,
"com.mendix.storage.s3.TokenService.RetryIntervalInSec": 10,
```

This means that the Mendix Runtime will proactively refresh the token at 80% of the expiry time of the current token. In case the TVM service is unreachable (think of maintenance) it has some time to refresh the token and it doesn't immediately mean that the Mendix Runtime doesn't have access to S3 anymore.

The `RetryIntervalInSec` setting means that in case it fails to connect to the TVM it will retry in 10 seconds instead of immediately.

buildpack/runtime_components/storage.py
pommi: There is a `util.get_buildpack_version()` function for this.

markbastiaans:
@manishjindal - thanks for your PR! Good to see that we're getting close to being able to use the TVM.

Be sure to:
- Add a clear description to your PR
- If possible / required, include unit / integration tests
- Ensure that your PR passes all checks / tests. At the moment, it is failing to lint - please use the `black` formatter with the supplied formatting rules to ensure that your code is "tidy"
- Rebase to the latest `develop` branch

buildpack/runtime_components/backup.py
```diff
@@ -22,17 +22,17 @@ def run():
         return
 
     backup_service = {}
-    if amazon_s3 in vcap_services:
-        s3_credentials = vcap_services[amazon_s3][0]["credentials"]
-        backup_service["filesCredentials"] = {
-            "accessKey": s3_credentials["access_key_id"],
-            "secretKey": s3_credentials["secret_access_key"],
```
pommi, line 29: `accessKey` and `secretKey` can be removed. The rest must stay.

buildpack/runtime_components/storage.py
```diff
@@ -53,6 +60,9 @@ def _get_s3_specific_config(vcap_services, m2ee):
 
     access_key = os.getenv("S3_ACCESS_KEY_ID", access_key)
     secret = os.getenv("S3_SECRET_ACCESS_KEY", secret)
+    tvm_endpoint = os.getenv("S3_TVM_ENDPOINT", tvm_endpoint)
+    tvm_username = os.getenv("S3_TVM_USERNAME", tvm_username)
+    tvm_password = os.getenv("S3_TVM_PASSWORD", tvm_password)
```
pommi, line 65: I think it's good to document these environment variables here as well: https://github.com/mendix/cf-mendix-buildpack#s3-settings

Infra-Red:
Hi @manishjindal and @pommi! Do you have any plans to publish API spec for Token Vending Machine? Currently, `com.mendix.storage.s3.tokenService.Url` property implements magic URL and it's not clear how to use it.

buildpack/runtime_components/storage.py
```diff
@@ -65,17 +75,63 @@ def _get_s3_specific_config(vcap_services, m2ee):
...
+                    "com.mendix.storage.s3.SecretAccessKey": secret,
+                    "com.mendix.storage.s3.BucketName": bucket,
+                }
+            else:
+                logging.error(
+                    "Failed to get AccessKeyID or SecretAccessKey from TVM"
+                )
+                return None
+        else:
+            logging.info("S3 config detected, activating external file store")
```
pommi, line 122: Could the text be altered a little, like "S3 TVM config detected, activating external file store"? Then we can see the difference between using the TVM and regular IAM credentials (on line 79).

pommi (approved):
Looks good to me :+1:

svanderburg (approved):

PR #421, opened 2021-04-26
Fix integration tests using database

markbastiaans:
This PR introduces a fix to account for integration tests that were failing due to an issue with retrieving the database port.

Specifically, this fix accounts for differences in the output of the `docker port` command used to retrieve information about the database container used in these tests.


pommi (approved):
Apart from the fact that the commit message doesn't tell me why the change was required, the change it self looks good.

PR #424, opened 2021-05-06
Adding support for optional InstrumentationConfig to MxAgent

sayyildiz:
This PR adds support to handle an optional custom env variable to pass an instrumentation config (JSON) to MxAgent


markbastiaans:
Hi @sayyildiz - thanks for your PR. Some questions / comments:

- Since the metrics agent instrumentation config is somewhat more complicated than the metrics agent config, is there a possibility to include a default instrumentation config that trims away the high-cardinality tags?
- Your code looks good. However, it fails in the linting stage. Can you ensure that it adheres to the included `black` formatting rules and passes `pylint`?

PR #425, opened 2021-05-12
Adding support for optional InstrumentationConfig to MxAgent

praveenbalu:
This PR adds support to handle an optional custom env variable to pass an instrumentation config (JSON) to MxAgent


praveenbalu:
Raising a new one

PR #432, opened 2021-05-18
Added default instrumentation config for fallback

praveenbalu:
Added default instrumentation config for fallback mechanism

### DIFF
--- a/buildpack/runtime.py
+++ b/buildpack/runtime.py
@@ -73,22 +73,6 @@ def stage(buildpack_dir, build_path, cache_path):
         os.path.join(build_path, ".local", "m2ee.yaml"),
     )
 
-    util.mkdir_p(os.path.join(build_path, ".local", "mx-agent"))
-    shutil.copy(
-        os.path.join(
-            buildpack_dir,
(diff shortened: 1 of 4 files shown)


buildpack/runtime.py
```diff
@@ -75,6 +75,22 @@ def stage(buildpack_dir, build_path, cache_path):
         os.path.join(build_path, ".local", "m2ee.yaml"),
     )
 
+    util.mkdir_p(os.path.join(build_path, ".local", "mx-agent"))
```
markbastiaans, line 78: Please don't add this to `runtime.py`, but to the staging phase in `mx-java-agent.py`.

markbastiaans (changes requested):
In general, LGTM, but please add the copying of the default config to the staging phase in `mx-java-agent.py`.

buildpack/runtime.py
praveenbalu: Done

carbone-kid:
Looks good to me.

carbone-kid (approved):

PR #434, opened 2021-05-19
v4.16.2: Graceful shutdown hotfix, instrumentation config

markbastiaans:
This release contains the following improvements:
- #424: add support for Mendix Java Agent instrumentation configuration 

This release also fixes an issue with graceful shutdown (1700d5c, #433, 344dad4). This fix should re-enable before-shutdown microflows. Note that depending on before-shutdown microflows is risky, since they won't be executed if the shutdown is not graceful.


ajouwens-mendix (approved):
Approved

PR #435, opened 2021-05-24
Storage S3 STS support for Mendix 7/8

pommi:
* Fetch STS token from TVM for Mendix >= 7.23.22
* Fetch STS token from TVM for Mendix >= 8.18.7
* Fix User-Agent header when fetching STS token from TVM

### DIFF
--- a/buildpack/runtime_components/storage.py
+++ b/buildpack/runtime_components/storage.py
@@ -5,6 +5,7 @@ import os
 import requests
 import time
 from buildpack import util
+from lib.m2ee.version import MXVersion
 
 
 def _get_s3_specific_config(vcap_services, m2ee):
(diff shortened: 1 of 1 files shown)


pommi:
@markbastiaans @mbyregowMx @svanderburg This PR is ready for review and can be merged.

@manishjindal Can you give your blessing as well?

ravi4mendix (approved):

markbastiaans:
Hi @pommi - can you rebase against the latest `develop`? Thanks!

buildpack/runtime_components/storage.py
```diff
@@ -91,21 +92,18 @@ def _get_s3_specific_config(vcap_services, m2ee):
...
         and tvm_password
-        and m2ee.config.get_runtime_version() < 9.2
-    ):
-        logging.info(
-            "S3 TVM config detected, fetching IAM credentials from TVM"
-        )
-        access_key, secret = _get_credentials_from_tvm(
-            tvm_endpoint, tvm_username, tvm_password
+        and (
+            m2ee.config.get_runtime_version() >= 9.2
```
markbastiaans, line 96: As an alternative, you can use `from buildpack import runtime` and `runtime.get_version()` to get the version. This will not require the m2ee client and get the version directly from metadata.

markbastiaans (changes requested):
LGTM, I added a comment which you are free to use (or not, of course).

Also, please rebase against the latest `develop`.

buildpack/runtime_components/storage.py
pommi: Thanks for your feedback! Done :+1:

pommi: Sorry @markbastiaans I just reverted it again. It seems `runtime.get_version()` requires an argument (build_path).

Turning:
```
m2ee.config.get_runtime_version()
```
into:
```
runtime.get_version(os.path.abspath("."))
```
doesn't make it more readable to me. Can the build_path be autodetected? `m2ee.config.get_runtime_version()` also doesn't require a build_path argument.

markbastiaans: Hi @pommi - I have an MR that will automate this. You can leave the code as-is and we can change it in the future.

markbastiaans (approved):

PR #439, opened 2021-05-28
Disable proxy buffering for files

markbastiaans:
This PR disables `nginx` proxy (request) buffering for the runtime file handler endpoint (`/file`).
Additionally, it refactors the location `proxy_intercept_errors` setting to match the proxy buffering setting.

---

Tested by sshing into the app container and verifying the Nginx config:
```
> cf ssh d0979a7c-0476-4df3-b984-a15085ce4132
vcap@bf5a202e-cf80-4d6a-5bee-8ed6:~$ cd app/
vcap@bf5a202e-cf80-4d6a-5bee-8ed6:~/app$ ./nginx/sbin/nginx -p nginx/ -c /home/vcap/app/nginx/conf/nginx.conf -T | grep -A 10 "location \/file"
nginx: the configuration file /home/vcap/app/nginx/conf/nginx.conf syntax is ok
nginx: configuration file /home/vcap/app/nginx/conf/nginx.conf test is successful
        location /file {
            if ($request_uri ~ ^/(.*\.(css|js)|forms/.*|img/.*|pages/.*)\?[0-9]+$) {
                    expires 1y;
            }
            proxy_pass http://mendix;
                    proxy_buffering off;
            proxy_request_buffering off;
            proxy_intercept_errors on;
            satisfy any;

        }
```

PR #442, opened 2021-06-21
Fix nginx error handling regression

markbastiaans:
This PR fixes a regression which was introduced in #439. This regression caused errors for some locations (e.g. `/rest`) to be handled by the `nginx` proxy instead of the Mendix runtime.

The fix makes the `proxy_intercept_errors` explicit for each configured location.


mbyregowMx (approved):
Changes are fine

PR #445, opened 2021-06-24
Fix Telegraf metrics submission to non-US Datadog orgs

markbastiaans:
This PR fixes Telegraf metrics submission for non-US Datadog orgs.


udh-mendix (approved):
LGTM

PR #448, opened 2021-07-06
Improve bootstrap logic

markbastiaans:
This PR includes the following changes:
* Remove Python wheels from git repository
* Include logic for online bootstrapping when wheels are not bundled
* Remove dependency on `.commit` file


bin/bootstrap-python
```diff
@@ -10,34 +10,43 @@ CACHE_PATH=$2
...
+        curl https://bootstrap.pypa.io/get-pip.py -o get-pip.py
+        python3 get-pip.py install $PIP_VERBOSITY_FLAGS --user --no-warn-script-location
+    fi
+    PIP_CMD="/home/vcap/.local/bin/pip"
+else
+    # pip is bundled on the root FS
+    pip install --upgrade --user --no-warn-script-location pip setuptools wheel
+fi
 
-if [[ $(ls -1q $PIP_WHEELDIR/*.whl | wc -l) = 2 ]]
```
svanderburg, line 40: I don't fully understand where this number 4 comes from (or 2). Is there a way to make this more clear in the code? I guess the number 4 comes from the fact that we need 4 dependencies?

markbastiaans: `pip`, `setuptools` and `wheel` are three packages, which are now included as bootstrap requirements. There can be a couple of options here:
- The three wheels are not bundled - full online mode (bootstrap with `get-pip.py` and download other requirements from PyPa) (`< 4` wheels in directory)
- The three wheels are bundled only - only download other requirements (`< 4` wheels in directory)
- The three wheels and all other requirements are bundled - don't download (` >= 4` wheels in directory)

markbastiaans: Added code comments to clarify

svanderburg (approved):

PR #453, opened 2021-07-20
fix #452 inline postgres certificates

### LINKED ISSUE #452: inline postgres certificates unsupported
I deploy a web app to a local cloudfoundry environment. As a database service for my DEV environment I have chosen a Marketplace service google-cloudsql-postgres with the plan postgres-db-f1-micro. Using the Web UI I created an instance with the name myapp-test-database and mentioned it in the CF Manifest:

    applications:
    - name: myapp-test
      services:
       - myapp-test-database

With this configuration the Mendix Runtime fails with:

    ERROR - ConnectionBus: Opening JDBC connection to Some(34.91.242.101) failed with SQLState: 08006 Error code: 0 Message: Could not open SSL certificate file -----BEGIN CERTIFICATE-----
    ...
    Caused by: java.io.FileNotFoundException: -----BEGIN CERTIFICATE-----
    ...
    -----END CERTIFICATE----- (File name too long)

It's because the JDBC driver does not support inline certificates in `DATABASE_URL=postgres://user:pass@host?sslcert=-----BEGIN+CERTIFICATE-inlineCertificate... `

Here: https://github.com/mendix/cf-mendix-buildpack/blob/ce60432790e501e4237cbeca62b1a33c4849a77c/buildpack/runtime_components/database.py#L347
the jdbc params need to be parsed further, the inline certificates extracted to disk and PKCS#1 PEM key converted to PKCS#8 binary key that Postgres driver understands.

basi01:
This PR adds inline certificate support to JDBC parameter parsing.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,7 +13,7 @@ download_wheels: requirements create_build_dirs
 	rm -rf build/vendor/wheels
 	mkdir -p build/vendor/wheels
 	pip3 download -d build/vendor/wheels/ --only-binary :all: pip setuptools wheel
-	pip3 download -d build/vendor/wheels/ --no-deps --platform manylinux1_x86_64 --python-version 36 -r requirements.txt
+	pip3 download -d build/vendor/wheels/ --no-deps --platform manylinux2010_x86_64 --python-version 36 -r requirements.txt
 
 .PHONY: create_build_dirs
 create_build_dirs:
--- a/buildpack/runtime_components/database.py
+++ b/buildpack/runtime_components/database.py
@@ -3,6 +3,8 @@ import logging
 import os
 import re
 import shutil
+import subprocess
+import tempfile
 from abc import ABC, abstractmethod
 from urllib.parse import parse_qs, unquote, urlencode
 
@@ -247,6 +249,12 @@ class DatabaseConfiguration(ABC):
 class UrlDatabaseConfiguration(DatabaseConfiguration):
     # Returns a database configuration based on the original code from util.
 
+    SSLROOTCERT = "sslrootcert"
+
+    SSLKEY = "sslkey"
+
+    SSLCERT = "sslcert"
+
     def __init__(self, url, env_vars=None):
         super().__init__(env_vars)
         logging.debug("Detected URL based database configuration.")
@@ -319,6 +327,11 @@ class UrlDatabaseConfiguration(DatabaseConfiguration):
             )
             jdbc_params.update({"sslmode": "verify-full"})
 
+        if database_type == "PostgreSQL":
+            self.extract_inline_cert(jdbc_params, self.SSLCERT, "postgresql.crt");
+            self.extract_inline_cert(jdbc_params, self.SSLKEY, "postgresql.pk8");
+            self.extract_inline_cert(jdbc_params, self.SSLROOTCERT, "root.crt");
+
         extra_url_params_str = self.env_vars.get(
             "DATABASE_CONNECTION_PARAMS", "{}"
         )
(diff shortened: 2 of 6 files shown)


markbastiaans:
> The first commit converts the private key using the openssl binary. The second commit uses python cryptography instead, but it pulls 4 more python dependencies. Which is better?

Thanks for your PR! To answer your direct question, we prefer Python `cryptography` since it abstracts away OpenSSL, enabling us to (unit) test this on different systems OSes.

Some comments about your PR:
- Please rebase on the latest `develop` branch
- Please ensure that our test suite can run by linting your code - use `make lint` to check locally if your code will pass
- Please use the most recent version of `cryptography` - the PR currently fails the Snyk security scan because it pulls in old(er) versions of its dependencies
- Please include (unit) test cases for the functionality you added

Thanks in advance for your improvements - looking forward to further review.

basi01:
> Some comments about your PR

done

basi01:
I don't know why, but `make build` downloads `cryptography-3.4.7.tar.gz` instead of `cryptography-3.4.7-cp36-abi3-manylinux2014_x86_64.whl` and packs it to `cf-mendix-buildpack.zip`. At the same time, `make install_requirements` installs `cryptography-3.4.7-cp36-abi3-manylinux2014_x86_64.whl` in the current docker app (dev_buildpack)

markbastiaans:
@basi01 - be sure to run `make requirements` to ensure that `requirements.txt` is up to date. At the moment, we cannot proceed with the review as the integration tests are failing for your PR.

Also, would it be possible to generate the certificates you use for testing on the fly, instead of including them in the repository?

basi01:
To make integration tests work I need to target the newer platform "manylinux2010_x86_64". Will you approve this? 
```
--- a/Makefile
+++ b/Makefile
@@ -13,7 +13,7 @@ download_wheels: requirements create_build_dirs
        rm -rf build/vendor/wheels
        mkdir -p build/vendor/wheels
        pip3 download -d build/vendor/wheels/ --only-binary :all: pip setuptools wheel
-       pip3 download -d build/vendor/wheels/ --no-deps --platform manylinux1_x86_64 --python-version 36 -r requirements.txt
+       pip3 download -d build/vendor/wheels/ --no-deps --platform manylinux2010_x86_64 --python-version 36 -r requirements.txt
```

basi01:
> would it be possible to generate the certificates you use for testing on the fly

I'll try to port the openssl commands to cryptography because I prefer to test in Visual Studio where openssl.exe is not in %PATH%. I hope this won't take much time.

basi01:
done

tests/unit/certtool.py
```diff
@@ -0,0 +1,100 @@
...
+# https://cryptography.io/en/latest/x509/tutorial/
+class CertGen:
+    def __init__(self):
+        self.init_root_cert()
+        self.init_postgresql_cert()
+        self.dump_to_storage()
+
+    def dump_to_storage(self):
+        self.cert_map = {}
+        self._dump_cert(self.root_cert, "root.crt")
```
markbastiaans, line 19: Can you perhaps use temp files for both files, and pass the paths / handles on to the consumer?
Currently, there might be leftovers when you run unit tests locally (so outside of a container).

tests/unit/certtool.py
```diff
@@ -0,0 +1,100 @@
+import datetime
```
markbastiaans, line 1: Since this class is only used in the inline cert test, it might be an option to move it to `tests/unit/test_db_config_options.py` instead of it having its own file.

markbastiaans (changes requested):
LGTM, aside from two minor comments.

tests/unit/certtool.py
basi01: The temporary files created by database.py are supposed to stay after the python program exits. The consumer can later extract the paths from the jdbc URL and delete them.

The unit test does not create additional temporary files. It stores the expected text and binary payloads in latin-1 strings in a dictionary. File-like names are only to distinguish different kinds of certificates in `cert_map`. latin-1 encoding is chosen because most of the time I compare PEM payloads, but it's good for binary too.

If the check succeeds the unit test deletes the temporary files created by database.py. If it fails, some files will remain, but maybe it's for the better as it makes troubleshooting easier. So I don't want to add the additional cleanup here.

tests/unit/certtool.py
basi01: Refactored...

markbastiaans (approved):
Thanks @basi01 - this one is ready to merge.

PR #459, opened 2021-07-30
Update Python dependencies

markbastiaans:
This PR updates all Python dependencies, and bumps the included wheel platform to `manylinux2014`. It also explicitly pins `urllib3` to mitigate CVE-2021-33503.

Additionally, the `black` and `pylint` line length has been aligned, and code has been formatted according to the new specifications introduced by newer versions of these dependencies.

Note that the CA certificate bundle used in this buildpack is now also upgraded to `certifi 2021.5.30`. Please upgrade, renew or switch certificates you might be using which are not included in this bundle.


mukund-padale (approved):
Changes are fine hence approving.

PR #471, opened 2021-09-23
v4.21.1

svanderburg:
This hotfix release contains the following changes:
* Disable S3 STS support when CAs are set (#469)


mbyregowMx (approved):
approved

PR #480, opened 2021-11-15
Add "+" to allowed Content-Security-Policy header

markbastiaans:
This PR introduces a fix for the allowed `Content-Security-Policy` header. It now allows including the `+` sign, which allows for `base64` strings in the header.


mbyregowMx (approved):
Changes looks fine, unit test is done. So approving the PR

PR #488, opened 2021-12-14
LM-134: Add file size metrics

jeohist:
In Mendix 7.4 we added a `size` column to `system$filedocument`
which allows us to get an overview of the file storage usage:
<https://docs.mendix.com/releasenotes/studio-pro/7.4#other-improvements>
This commit outputs these metrics for compatible versions.


joshp123:
This may have product considerations, let's verify this + get sign-off from them before merging.

tomedegroot:
Hi Joey,

Great that you take the initiative. I guess we need to work on the CloudPortal side as wel to actually display the graphs to users?

buildpack/telemetry/metrics.py
```diff
@@ -484,21 +493,13 @@ def _get_number_of_files(self):
...
-                    return 0
-                return int_or_default(rows[0][0])
-            except Exception:
-                # We ignore errors here, as the information is
-                # not available for older mendix versions
-                logging.debug(
-                    "METRICS: Error retrieving file sizes", exc_info=True
-                )
-                return 0
+            cursor.execute(
```
udh-mendix, line 496: why remove cursor.execute from the try block?

jeohist: It's not removed, the try block is just moved to L353, to be consistent with the way we get the number of files. The old logic essentially went "try to get size of files, if it fails, must be the version!". In this way it first checks the version, and then if it fails it'll throw an actual exception.

udh-mendix (approved):
LGTM

PR #499, opened 2022-02-01
Nginx should cache static client images

Robinfr:
The upcoming version of Mendix will have static images in a separate folder and should be cached in order for apps to continue working as expected.

### DIFF
--- a/etc/nginx/conf/nginx.conf.j2
+++ b/etc/nginx/conf/nginx.conf.j2
@@ -93,7 +93,7 @@ http {
         {%- endmacro %}
         {% for location in locations %}
         location {{ location.path }} {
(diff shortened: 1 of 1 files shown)


markbastiaans:
Hi @Robinfr - thanks for your contribution!

I've changed the target branch of this PR to `develop` for you. In this way, it will be included in our regular release cycle.

Although this change is small, the implications may be big. Can you provide us with more information about:
- How you tested this change
- Whether or not this change is safe, i.e. backward-compatible for earlier Mendix versions

Ideally, changes like these are accompanied with test cases. In this way, we can somewhat guarantee expected behaviour in the future. It would be great if you could include test cases if possible. There are some integration test cases already that deal with custom headers and access restrictions, maybe this could be part of these.

markbastiaans (approved):

PR #481, opened 2021-11-17
Switch to metrics via micrometer

pappachino:
Starting MxRuntime v9.7.0, runtime metrics will be fetched via micrometer instead of admin port.

This PR includes changes to :
- Switch off runtime application statistics from admin port, if metrics via micrometer is enabled
- Bump telegraf metric buffer limit from 10k -> 15k to address metric buffer overflow
- For freeapps, configure micrometer to push only session metrics

Known Issue:
---
Few runtime metrics is displayed as `NaN` in the trends graph if the action the metric represents has not been invoked yet. This affects MxRuntime v9.7.x and v9.8.x. This issue is fixed in MxRuntime v9.9.0

PR #500, opened 2022-02-01
Hotfix : Fix regression to enable database metrics in telegraf only when needed

pappachino:
With micrometer metrics enabled by default for all apps (#481), database metrics related telegraf input plugin was enabled. This introduced a regression, as now for all apps we started running postgres queries. This change fixes the regression.

Going forward,
- For freeapps, we never include any database metrics via telegraf.
- For paidapps, any customers who have set DD_API_KEY or APPMETRICS_TARGET would get the database metrics via telegraf.


buildpack/telemetry/telegraf.py
```diff
@@ -59,7 +59,18 @@ def _get_appmetrics_target():
...
+    if util.is_free_app():
+        # For free apps we are not interested in database metrics
+        return False
+
+    value = "false"
+    if _get_appmetrics_target() is not None or datadog.is_enabled():
+        # For customers who have Datadog or APPMETRICS_TARGET enabled,
+        # we always include the database metrics. They can opt out
+        # using the APPMETRICS_INCLUDE_DB flag
+        value = "true"
```
udh-mendix, line 71: I think its better if the APPMETRICS_INCLUDE_DB check is inside this if block, so at the end you can simply return strtobool(value)

udh-mendix:
The MR is not only about free apps, it fixes the regression which enabled database metrics in telegraf for ALL apps. Would be nice if its made clear in the MR description and title, thanks!

buildpack/telemetry/telegraf.py
pappachino: I've included the env var check inside the block now.

udh-mendix (commented):
LGTM

udh-mendix (commented):
LGTM

ajinkya-takle (dismissed):
LGTM!

svanderburg (approved):

PR #507, opened 2022-03-08
COA-735: Update storage deletion behaviour

jeohist:
This also relates to RTO-3039.

The behaviour of `PerformDeleteFromStorage` changed in 9.6.10 / 9.12.
There are now three options:
* NoFiles
* TemporaryFiles (formerly False)
* AllFiles (formerly True)

`dont_perform_deletes` implies no deletes, which means
that NoFiles is the appropriate option from this point forward.

Since this will be released with a new MTS release
we'll bump that part of the code while we're at it.

---

Also bumps dependencies:

* Bump certifi (only adds an additional certificate)
* Bump httplib2 / remove urllib3 (mitigation in place)
* Bump others (just generally a good thing to do?)

Relying on automated tests to ensure stuff keeps working there ;-)


requirements.txt
```diff
@@ -1,5 +1,5 @@
 #
-# This file is autogenerated by pip-compile with python 3.6
+# This file is autogenerated by pip-compile with python 3.9
```
markbastiaans, line 2: Don't know if the dependency bumps will be an issue since you didn't use a Python 3.6 environment to generate the dependency matrix - did you test this?

markbastiaans (changes requested):

requirements.txt
jeohist: I did test this but forgot to re-do the pip-compile with 3.6, lemme fix it

buildpack/infrastructure/storage.py
```diff
@@ -134,6 +134,13 @@ def _get_s3_specific_config(vcap_services):
         if runtime.get_runtime_version() < 7.19:
             # Deprecated in 7.19
             config[config_prefix + "PerformDeleteFromStorage"] = False
+        elif runtime.get_runtime_version() >= (
+            MXVersion("9.6.11") or MXVersion("9.12")
```
markbastiaans, line 138: Not really familiar with how Python does it, but this rings a bit of an alarm bell to me - in other languages, this would be equal to comparing one value to whether one of the other values is true (which is always the case here).

jeohist: I've made it consistent with the way that we compare versions in the rest of the code now.

markbastiaans (approved):

PR #509, opened 2022-03-15
Disable S3 STS support when CAs are set for Mendix 9.6.0

pommi:
In #504 this feature was enabled for Mendix >= 9.6, but the actual fix is in 9.6.1.


markbastiaans (changes requested):
Since this is a hotfix, please do a version bump in `VERSION` so that we can directly merge to `master`.

markbastiaans (approved):

PR #512, opened 2022-03-17
v4.26.3: Fix user signal handling

markbastiaans:
This PR attempts to fix user signal handling for `SIGUSR1`and `SIGUSR2`. These signals will now cause a `sys.exit(1)` even if the Mendix metrics module is disabled.


tests/integration/test_termination.py
```diff
@@ -2,33 +2,43 @@
...
-            self.assert_string_in_recent_logs(
-                "Runtime process has been terminated"
-            )
-        self.assert_string_in_recent_logs("Terminating process group")
-
-    # Tests if termination works if a shutdown command is sent to the runtime
-    def test_termination_shutdown_command(self):
-        self.stage_container("Mendix8.1.1.58432_StarterApp.mda")
+    def _test_termination_signal(
+        self, signal="SIGTERM", env_vars={}, exitcode=0
```
pappachino, line 6: Let's have `env_vars` default to None, else we might run into the mutable default argument gotcha.

pappachino (approved):
Looks good!!

svanderburg (approved):

PR #514, opened 2022-03-28
Fix regression for Metrics.Registries custom runtime setting

pappachino:
With metrics via micrometer, we should be able to configure any registries as detailed in the [Mendix docs](https://docs.mendix.com/refguide/metrics/#registries-configuration). 

However, a regression was causing additional registries like prometheus from being added. This PR fixes the regression.


ekremsekerci (approved):
all looks good, also completed peer testing.

markbastiaans (approved):

PR #516, opened 2022-03-31
feat(finalize): add launch environment support

Infra-Red:
Make sure that `profile.d` scripts that are provided by other (`supply`) buildpacks are copied by the last (`finalize`) buildpack [1].

During startup, the container starts a bash shell that source’s all `.sh` scripts in the `.profile.d/` directory before executing the app’s command [2]. Such scripts can be provided by other buildpacks and the last buildpack may make use of dependencies by copying files to the build directory.

Changes that are added by this PR are Python implementation of the same functionality from the "official" buildpack helper library:
- https://github.com/cloudfoundry/libbuildpack/blob/2060a8f734b232f08362a1856c1993cb05072be0/stager.go#L259-L314

[1] https://docs.cloudfoundry.org/buildpacks/custom.html#contract
[2] https://devcenter.heroku.com/articles/buildpack-api#profile-d-scripts

### DIFF
--- a/buildpack/util.py
+++ b/buildpack/util.py
@@ -237,6 +237,43 @@ def get_mpr_file_from_dir(directory):
         return None
 
 
+def set_up_launch_environment(deps_dir, profile_dir):
+    profile_dirs = get_existing_deps_dirs(deps_dir, "profile.d", deps_dir)
+    for dir in profile_dirs:
+        sections = dir.split(os.sep)
+        if len(sections) < 2:
+            raise Exception("Invalid dependencies directory")
+
+        deps_idx = sections[len(sections) - 2]
+
+        files = [
+            f for f in os.listdir(dir) if os.path.isfile(os.path.join(dir, f))
+        ]
+
+        for file in files:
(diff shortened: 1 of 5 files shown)


buildpack/util.py
```diff
@@ -237,6 +237,43 @@ def get_mpr_file_from_dir(directory):
         return None
 
 
+def set_up_launch_environment(deps_dir, profile_dir):
+    profile_dirs = get_existing_deps_dirs(deps_dir, "profile.d", deps_dir)
```
markbastiaans, line 241: Should this be `.profile.d` instead of `profile.d`?

Infra-Red: The directory name in buildpack dependenices is `profile.d`, and it gets copied to `.profile.d` build folder.

From the [docs](https://docs.cloudfoundry.org/buildpacks/custom.html#contract):
> To make use of dependencies provided by previous buildpacks, the last buildpack:
> * May copy files from /profile.d to /tmp/app/.profile.d during staging.

markbastiaans (approved):

PR #523, opened 2022-04-20
Hotfix v4.28.1: Update telegraf config to ignore metrics from influxdb listener plugin

pappachino:
If APPMETRICS_TARGET is set, we use statsd input telegraf plugin. Ignore any metrics originating from influxdb listener plugin.


udh-mendix (approved):
LGTM

PR #525, opened 2022-04-22
Enable S3 STS support when CAs are set for Mx7

nityashandilya:
The issue which we made a workaround for in https://github.com/mendix/cf-mendix-buildpack/pull/469 is also fixed in Mendix 7.23.30.

### DIFF
--- a/buildpack/infrastructure/storage.py
+++ b/buildpack/infrastructure/storage.py
@@ -161,8 +161,10 @@ def _get_s3_specific_config(vcap_services):
 
 
 def _runtime_sts_support(version):
-    if version >= MXVersion("9.6.1") or (
-        version.major == 8 and version >= MXVersion("8.18.11")
+    if (
+        version >= MXVersion("9.6.1")
+        or (version.major == 8 and version >= MXVersion("8.18.11"))
+        or (version.major == 7 and version >= MXVersion("7.23.30"))
     ):
         return True
     # Only enable STS support for these versions when CERTIFICATE_AUTHORITIES


buildpack/infrastructure/storage.py
```diff
@@ -170,7 +170,7 @@ def _runtime_sts_support(version):
     elif (
         version >= 9.2
         or (version.major == 8 and version >= MXVersion("8.18.7"))
-        or (version.major == 7 and version >= MXVersion("7.23.22"))
+        or (version.major == 7 and version >= MXVersion("7.23.30"))
```
pommi, line 173: This check should not change.

The previous `if` statement where we check for `>= 9.6.1` and `>= 8.18.11` requires an additional statement where we check for `>= 7.23.30` as well.

tests/unit/test_s3_blobstore.py
```diff
@@ -172,7 +172,7 @@ def test_s3_blobstore_tvm_runtime_with_sts(self):
 
     """
     STS support is broken when CERTIFICATE_AUTHORITIES is set for Mendix
-    versions 7.23.22-, 8.18.7-8.18.10 and 9.2-9.5). Fallback to IAM
+    versions 7.23.30-, 8.18.7-8.18.10 and 9.2-9.5). Fallback to IAM
```
pommi, line 175: This shouldn't change as well.

pommi (changes requested):

buildpack/infrastructure/storage.py
nityashandilya: I did the changes, could you please check that.

pommi (approved):

PR #528, opened 2022-04-25
LM-774 default env vars for AppDynamics

denis-shuvalov:
Added default values for environment variables for AppDynamics. 
If the variables are not set up by user, it will be defined by the application features:

```
        "APPDYNAMICS_AGENT_APPLICATION_NAME": App name from Cloud Portal,
        "APPDYNAMICS_AGENT_NODE_NAME": "node",
        "APPDYNAMICS_AGENT_TIER_NAME": Environment uuid,
        "APPDYNAMICS_CONTROLLER_PORT": "443",
        "APPDYNAMICS_CONTROLLER_SSL_ENABLED": "true",
```

Fixed log messages for AppDynamics.

### DIFF
--- a/README.md
+++ b/README.md
@@ -404,25 +404,28 @@ To enable AppDynamics, configure the following environment variables.
 These settings enable basic (Java Agent related) AppDynamics features.
 More information here: [Environment variables](https://docs.appdynamics.com/22.2/en/application-monitoring/install-app-server-agents/java-agent/install-the-java-agent/use-environment-variables-for-java-agent-settings#UseEnvironmentVariablesforJavaAgentSettings-EnvironmentVariables).
 
-Please note that AppDynamics requires Mendix 6.2 or higher.
+Please note that AppDynamics requires Mendix 7.15 or higher.
 
-| Environment Variable | Value Example | Description |
-|-|-|-|
-| `APPDYNAMICS_CONTROLLER_PORT` | 443 | Port of AppDynamics controller |
-| `APPDYNAMICS_CONTROLLER_SSL_ENABLED` | true | Set if SSL is required |
-| `APPDYNAMICS_CONTROLLER_HOST_NAME` | account-name.saas.appdynamics.com | Name of AppDynamics host without 'http://' |
-| `APPDYNAMICS_AGENT_APPLICATION_NAME` | myapp | Name of the app to be displayed on Controller UI |
-| `APPDYNAMICS_AGENT_ACCOUNT_NAME` | account-name | See 'License/Account' on the Controller UI |
-| `APPDYNAMICS_AGENT_ACCOUNT_ACCESS_KEY` | secret_key | See 'License/Account' on the Controller UI |
-| `APPDYNAMICS_AGENT_NODE_NAME` * | mynode | How a node is displayed on the Controller UI |
-| `APPDYNAMICS_AGENT_TIER_NAME` | mytier | How a tier is displayed on the Controller UI |
+| Environment Variable | Value example | Default | Description |
+|-|-|-|-|
+| `APPDYNAMICS_CONTROLLER_PORT` | `443` | `443` | Port of AppDynamics controller |
+| `APPDYNAMICS_CONTROLLER_SSL_ENABLED` | `true` | `true` | Set if SSL is required |
+| `APPDYNAMICS_CONTROLLER_HOST_NAME` | `<acc_name>.test.com` | - | Name of AppDynamics host without 'http://' |
+| `APPDYNAMICS_AGENT_APPLICATION_NAME` | `<test-accp>` | App name as on Dev. Portal | Name of the app to be displayed on Controller UI |
+| `APPDYNAMICS_AGENT_ACCOUNT_NAME` | `<acc_name>` | - | See 'License/Account' on the Controller UI |
+| `APPDYNAMICS_AGENT_ACCOUNT_ACCESS_KEY` | `<secret>` | - | See 'License/Account' on the Controller UI |
+| `APPDYNAMICS_AGENT_NODE_NAME` * | `<node>` (finally `<node>_<num>`, `<num>` is being added automatically) | node | How a node is displayed on the Controller UI |
+| `APPDYNAMICS_AGENT_TIER_NAME` | `<env_id>` | App Environment UUID | How a tier is displayed on the Controller UI |
 
 
-\* The `APPDYNAMICS_AGENT_NODE_NAME` environment variable will be appended with the value of the `CF_INSTANCE_ID` variable. If you use `mynode` for `APPDYNAMICS_AGENT_NODE_NAME` , the AppDynamics agent will be configured as `mynode-0` for instance `0` and `mynode-1` for instance `1` , etc.
+\* The `APPDYNAMICS_AGENT_NODE_NAME` environment variable will be appended with the value of the `CF_INSTANCE_ID` variable. If you use `node` for `APPDYNAMICS_AGENT_NODE_NAME` , the AppDynamics agent will be configured as `node_0` for instance `0` and `node_1` for instance `1` , etc.
 
 For more details about nodes and tiers: [Tiers and Nodes](https://docs.appdynamics.com/22.1/en/application-monitoring/tiers-and-nodes).
 
-If you have all the environment variables specified above, the AppDynamics Java Agent will be configured for your application. 
+Required variables: `APPDYNAMICS_AGENT_ACCOUNT_NAME`, `APPDYNAMICS_AGENT_ACCOUNT_ACCESS_KEY` and `APPDYNAMICS_CONTROLLER_HOST_NAME`.
+
+If you have all the required environment variables specified above, the AppDynamics Java Agent will be configured for your application.
+The other environment variables are optional and will be set to the default values.
 After configuring these environment variables, restart your app for the agent to be enabled.
 If the agent has not been installed it is necessary to redeploy the app.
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -398,23 +398,26 @@ More information here: [Environment variables](https://docs.appdynamics.com/22.2
...
-| `APPDYNAMICS_AGENT_APPLICATION_NAME` | myapp | Name of the app to be displayed on Controller UI |
-| `APPDYNAMICS_AGENT_ACCOUNT_NAME` | account-name | See 'License/Account' on the Controller UI |
-| `APPDYNAMICS_AGENT_ACCOUNT_ACCESS_KEY` | secret_key | See 'License/Account' on the Controller UI |
-| `APPDYNAMICS_AGENT_NODE_NAME` * | mynode | How a node is displayed on the Controller UI |
-| `APPDYNAMICS_AGENT_TIER_NAME` | mytier | How a tier is displayed on the Controller UI |
+| `APPDYNAMICS_CONTROLLER_HOST_NAME` | - | Name of AppDynamics host without 'http://' |
+| `APPDYNAMICS_AGENT_APPLICATION_NAME` | App name as on Dev. Portal | Name of the app to be displayed on Controller UI |
+| `APPDYNAMICS_AGENT_ACCOUNT_NAME` | - | See 'License/Account' on the Controller UI |
+| `APPDYNAMICS_AGENT_ACCOUNT_ACCESS_KEY` | - | See 'License/Account' on the Controller UI |
+| `APPDYNAMICS_AGENT_NODE_NAME` * | node | How a node is displayed on the Controller UI |
```
pappachino, line 409: Would `node` be the hostname ? Maybe an example would help.

README.md
```diff
@@ -398,23 +398,26 @@ More information here: [Environment variables](https://docs.appdynamics.com/22.2
...
 
 \* The `APPDYNAMICS_AGENT_NODE_NAME` environment variable will be appended with the value of the `CF_INSTANCE_ID` variable. If you use `mynode` for `APPDYNAMICS_AGENT_NODE_NAME` , the AppDynamics agent will be configured as `mynode-0` for instance `0` and `mynode-1` for instance `1` , etc.
 
 For more details about nodes and tiers: [Tiers and Nodes](https://docs.appdynamics.com/22.1/en/application-monitoring/tiers-and-nodes).
 
-If you have all the environment variables specified above, the AppDynamics Java Agent will be configured for your application. 
+Required variables: `APPDYNAMICS_AGENT_ACCOUNT_NAME`, `APPDYNAMICS_AGENT_ACCOUNT_ACCESS_KEY` and `APPDYNAMICS_CONTROLLER_HOST_NAME`.
+
+If you have all the required environment variables specified above, the AppDynamics Java Agent will be configured for your application.
+The other variables (not required) will be set to the default values.
```
pappachino, line 420:
```suggestion
The other environment variables are optional and will be set to the default values.
```

buildpack/telemetry/appdynamics.py
```diff
@@ -18,6 +18,63 @@
...
+APPDYNAMICS_DEFAULT_ENV_VARS = {
+    "APPDYNAMICS_AGENT_APPLICATION_NAME": util.get_app_from_domain(),
+    APPDYNAMICS_NODE_NAME_KEY: "node",  # This is 'base' name without suffix
+    "APPDYNAMICS_AGENT_TIER_NAME": APPDYNAMICS_TIER,
+    "APPDYNAMICS_CONTROLLER_PORT": "443",
+    "APPDYNAMICS_CONTROLLER_SSL_ENABLED": "true",
+}
+
+
+def _filter_default_env():
```
pappachino, line 37: Falling back to a default value is already provided by `os.getenv`. Do we need this function ?

buildpack/telemetry/appdynamics.py
```diff
@@ -18,6 +18,63 @@
...
+        if var_name == APPDYNAMICS_NODE_NAME_KEY:
+            continue
+
+        if not (var_name in os_env_dict):
+            filtered_default_env_vars[var_name] = value
+
+    return filtered_default_env_vars
+
+
+FILTERED_APPDYNAMICS_DEFAULT_ENV_VARS = _filter_default_env()
```
pappachino, line 71: Why do we need to assign it to a variable and from the [all caps style](https://peps.python.org/pep-0008/#constants) it suggests it's a constant, which it is not.

README.md
denis-shuvalov: Added a description and example column

README.md
denis-shuvalov: done

buildpack/telemetry/appdynamics.py
denis-shuvalov: Good point, changed

buildpack/telemetry/appdynamics.py
denis-shuvalov: Doesn't exist anymore

buildpack/telemetry/appdynamics.py
```diff
@@ -18,6 +18,38 @@
     "machine-agent",
 )
 
+APPDYNAMICS_NODE_NUMBER = os.getenv("CF_INSTANCE_INDEX", default=0)
+APPDYNAMICS_TIER = util.get_vcap_data()["application_name"]
+
+# Number suffix will be added to AppDynamics Node Name
+# The key will be used in the function '_filter_default_env'
+APPDYNAMICS_NODE_NAME_KEY = "APPDYNAMICS_AGENT_NODE_NAME"
```
pappachino, line 26: Do we need this anymore ?

buildpack/telemetry/appdynamics.py
```diff
@@ -18,6 +18,38 @@
...
+
+APPDYNAMICS_DEFAULT_ENV_VARS = {
```
pappachino, line 28: Lets drop the `default`
```suggestion
APPDYNAMICS_ENV_VARS = {
```

pappachino (commented):
Looks good to me. Some minor comments.

buildpack/telemetry/appdynamics.py
denis-shuvalov: done

udh-mendix:
@denis-shuvalov do we need to update the mendix documentation as well?

pappachino (approved):
Looks good!

README.md
```diff
@@ -406,23 +406,26 @@ More information here: [Environment variables](https://docs.appdynamics.com/22.2
 
 Please note that AppDynamics requires Mendix 6.2 or higher.
```
markbastiaans, line 407: The buildpack only supports Mendix >= 7.

buildpack/telemetry/appdynamics.py
```diff
@@ -18,6 +18,40 @@
     "machine-agent",
 )
 
+CF_APPLICATION_INDEX = os.getenv("CF_INSTANCE_INDEX", default=0)
+CF_APPLICATION_NAME = util.get_vcap_data()["application_name"]
```
markbastiaans, line 22: This gives you the environment ID, which is not necessarily interesting to customers. Is this by design?

markbastiaans (changes requested):
A general remark: the AppDynamics integration tests are taking a long time to complete (in fact, the longest of all tests). This is causing every PR / release to be delayed more than it should be, IMO. If you can fix at least part of that in this PR, that would be great.

Quick wins:
- Remove the smoke test for Mendix 6. We don't support that version in the buildpack. This will already shave off 3-4 minutes per test run.
- Remove separate smoke test for "is the machine agent running" and include it in another smoke test

Longer-term wins:
- Please consider if you really have to start a full container to test functionality, or if you can cover what you're testing with unit tests. Examples are setting environment variables and other configuration settings - you can test expected outcomes with unit tests, and only run one smoke test to see if the agent(s) actually run or not.

buildpack/telemetry/appdynamics.py
denis-shuvalov: @markbastiaans The environment ID is provided as a "tier name". From the AppDynamics perspective, the "application name" is application name from domain, and that is the same as on Developer Portal, which makes sense for a user. Also, we consider "environment ID" as a logical value for "tier name" since it reflects the hierarchy on Developer portal. Also  user can find environment ID on Developer Portal.

Example hierarchy: "test application" / <environment id> / node_0

denis-shuvalov: Screenshot 2022-05-05 at 14 38 24

README.md
denis-shuvalov: Changed to 7.15. As in main documentation

denis-shuvalov:
> A general remark: the AppDynamics integration tests are taking a long time to complete (in fact, the longest of all tests). This is causing every PR / release to be delayed more than it should be, IMO. If you can fix at least part of that in this PR, that would be great.
> 
> Quick wins:
> 
> * Remove the smoke test for Mendix 6. We don't support that version in the buildpack. This will already shave off 3-4 minutes per test run.
> * Remove separate smoke test for "is the machine agent running" and include it in another smoke test
> 
> Longer-term wins:
> 
> * Please consider if you really have to start a full container to test functionality, or if you can cover what you're testing with unit tests. Examples are setting environment variables and other configuration settings - you can test expected outcomes with unit tests, and only run one smoke test to see if the agent(s) actually run or not.

@markbastiaans Mendix 6 tests removed. For test case "machine_agent_not_started" separate deploy is required becouse of different env. variables value, so it can not be included in any of the "positive" tests.

PR #551, opened 2022-05-25
LM-850 AppDynamics node name fix

denis-shuvalov:
Naming format for NODE is changed:

Underscore replaced for dash according to AppDynamics naming convention.

Old format: node_0
New format: node-1

The same change is relevant to Unique Host Id name.


pappachino (approved):
LGTM!

PR #556, opened 2022-06-08
Allow statsd_port to be configured through env. variables

Christophe-Vico:
While most of the Datadog env variables were supported, it was impossible to set the port for Datadog Statds.
The port was hardcoded in /buildpack/telemetry/datadog.py.

Change: 
- The value is now retrieved from the ENV Variable DD_DOGSTATSD_PORT if set, otherwise will be set to the default value 8125.

No change in the docs, it was already mentioned
"Other environment variables can be set as per the [Datadog Agent documentation](https://docs.datadoghq.com/agent/)."

Reasons for change:
- Statds does not start when the container in run on Fargate:
`2022-06-15 06:29:50 UTC | CORE | ERROR | (pkg/dogstatsd/server.go:247 in NewServer) | can't listen: listen udp 127.0.0.1:8125: bind: address already in use`
(while Datadog  will start just fine in a docker-compose setup)

- Silently overwriting DD_DOGSTATSD_PORT is also contradicting the documentation, as DD_DOGSTATSD_PORT is a Datadog standard config variable

### DIFF
--- a/buildpack/telemetry/datadog.py
+++ b/buildpack/telemetry/datadog.py
@@ -182,7 +182,7 @@ def _get_datadog_tags(model_version):
 
 
 def get_statsd_port():
-    return STATSD_PORT
+    return os.getenv("DD_DOGSTATSD_PORT", STATSD_PORT)
 
 
 def _set_up_dd_java_agent(


denis-shuvalov (approved):
LGTM

PR #558, opened 2022-06-30
Fix postgres query to collect temp_files stat

denis-shuvalov:
This MR adds a fix to address the incorrect telegraf postgres query to collect temp_files when datadog is enabled. This issue was noticed for datadog users that had `DATADOG_DATABASE_RATE_COUNT_METRICS` environment variable enabled.

Additionally, adds an environment variable to make the postgres metrics collection interval via telegraf configurable.


udh-mendix (approved):
LGTM

svanderburg (approved):

PR #560, opened 2022-07-12
Added trends forwarder support

denis-shuvalov:
Micrometer metrics are pushed to trends-forwarder service instead of trends-storage-server if corresponding feature flag is set.

### DIFF
--- a/buildpack/telemetry/metrics.py
+++ b/buildpack/telemetry/metrics.py
@@ -128,6 +128,26 @@ def get_appmetrics_target():
     return os.getenv("APPMETRICS_TARGET")
 
 
+def get_micrometer_metrics_url():
+    """
+    This function is used to provide selection of the trends storage URL.
+    There are two options - URL of trends-storage-server (old trends stack),
+    or URL of trends-forwarder (new trends stack). This selection is relevant for
+    micrometer metrics only. Runtime version 9.7 and above is required.
+
+    """
+    use_trends_forwarder = strtobool(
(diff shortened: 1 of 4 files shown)


buildpack/telemetry/metrics.py
```diff
@@ -128,6 +128,18 @@ def get_appmetrics_target():
     return os.getenv("APPMETRICS_TARGET")
 
 
+def get_micrometer_metrics_url():
```
pappachino, line 131: Few points:
- `USE_TRENDS_FORWARDER` as a feature flag is good, but we need to switch to `TRENDS_FORWARDER_URL` only if it is not empty.
- Can we add some comments or docstring to explain why we have a separate function as opposed to `get_metrics_url` below?
- How about we call this method `get_metrics_url` instead of having reference to the micrometer and have the other one be called `get_trends_storage_url` so it's unambiguous?
- Can we add a unit-test for this method or update any existing ones?

denis-shuvalov: @pappachino Another suggestion: we can leave `get_metrics_url` only, but add a parameter `micrometer_only=True', and call it this way for Telegraf

denis-shuvalov: * Added check for TRENDS_FORWARDER_URL 
* Docstring has been added
* Decided to keep names the same
* Unit test added

pappachino (approved):
LGTM!

svanderburg (approved):

PR #567, opened 2022-08-11
configure max_metaspace_size as env variable

geoffreymersch:
We may want to add the possibility to manually change the XX:MaxMetaspaceSize value for some customers.

### DIFF
--- a/buildpack/core/java.py
+++ b/buildpack/core/java.py
@@ -305,6 +305,10 @@ def _set_jvm_memory(m2ee, vcap):
     heap_size = str(heap_size) + "M"
 
     env_heap_size = os.environ.get("HEAP_SIZE")
+    max_metaspace_size = os.getenv("MAX_METASPACE_SIZE", "256M")
+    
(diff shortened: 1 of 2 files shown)


buildpack/core/java.py
```diff
@@ -305,6 +305,13 @@ def _set_jvm_memory(m2ee, vcap):
     heap_size = str(heap_size) + "M"
 
     env_heap_size = os.environ.get("HEAP_SIZE")
+    max_metaspace_size = os.environ.get("MAX_METASPACE_SIZE")
+    
+    if max_metaspace_size:
+        util.upsert_javaopts(m2ee, "-XX:MaxMetaspaceSize=%s" % max_metaspace_size)
+    else:
+        util.upsert_javaopts(m2ee, "-XX:MaxMetaspaceSize=256M")
+
```
svanderburg, line 314: This code fragment can be changed into a single line by using the `os.getenv` function: https://docs.python.org/3/library/os.html#os.getenv. The second parameter can be used to assign a default value if the environment variable is unspecified

svanderburg (approved):

PR #570, opened 2022-08-22
Handle database connection errors gracefully during stats gathering

pappachino:
For basic apps using Aurora serverless, the database connections are closed every day. 

From the [AWS docs](https://docs.aws.amazon.com/AmazonRDS/latest/AuroraUserGuide/aurora-serverless.html#aurora-serverless.limitations)
> Connections to an Aurora Serverless v1 DB cluster are closed automatically if held open for longer than one day.

So handle the db connections errors gracefully and don't raise connection errors. For the next metrics collection interval, the database connection would be refreshed and it would proceed to collect the relevant database stats.


ajinkya-takle (approved):
Changes looks good & it's peer tested in LM-868

PR #579, opened 2022-11-03
Update mendix-logfilter dependency to v0.0.2

pappachino:
This fixes the log truncation issue with log messages that start with digits.

For instance, before the fix texts like:
`123 : some message` ends up as `some message` and we lose the initial digits.


ekremsekerci (approved):
Tested it and looks good.

PR #582, opened 2022-11-21
Update micrometer registry to emit login metrics

ajinkya-takle:
Update micrometer registry to emit login metrics.

It's disabled by default & needs to be whitelisted explicitly. This metrics is enabled for paid as well as free apps. It's disabled by design to save ingestion costs when an external monitoring system is enabled on the Mendix app. This metric is mostly for the internal consumption.


pappachino (approved):
LGTM

PR #583, opened 2022-11-25
WIP: Metrics ingestion to Dynatrace

ekremsekerci:
* Remove AppMon agent since it's no longer supported https://www.dynatrace.com/support/appmon/
* Add telegraf output for dynatrace with url & token so that no agent needed
* Update readme accordingly
* Update telegraf version to  currrent latest version which is 1.24.4. Had to do to have default_dimensions attribute of dynatrace output plugin
* Use TAGS env var to push extra tags

### DIFF
--- a/README.md
+++ b/README.md
@@ -720,8 +720,7 @@ The most important metrics ( `before_xid_wraparound` , `connections` , `database
 
 ### Dynatrace
 
-[Dynatrace SaaS/Managed](http://www.dynatrace.com/cloud-foundry/) is your full stack monitoring solution - powered by artificial intelligence. Dynatrace SaaS/Managed allows you insights into all application requests from the users click in the browser down to the database statement and code-level.
-
+Dynatrace integration is supported in Mendix version 9.7 and above.
 To enable Dynatrace, configure the following environment variables:
 
 | Environment Variable | Description                                                                                   |
@@ -729,7 +728,12 @@ To enable Dynatrace, configure the following environment variables:
 | `DT_PAAS_TOKEN`      | The token for integrating your Dynatrace environment with your Mendix app                     |
 | `DT_SAAS_URL`        | Monitoring endpoint url of the Dynatrace service                                              |
 
-By setting these environment variables, ingestion of metrics to Dynatrace will start.
+Buildpack attaches these default tags to metrics that are pushed to Dynatrace:
+
+* `app` - Environment Id of the Mendix application
+* `instance_index` - Instance index that metrics belong to
+
+Extra tags can be attached via `TAGS` environment variable (example value: `[env:accp]`)
 
 ### Logging
 The buildpack provides several options to configure logging.
--- a/README.md
+++ b/README.md
@@ -724,15 +724,12 @@ The most important metrics ( `before_xid_wraparound` , `connections` , `database
 
 To enable Dynatrace, configure the following environment variables:
 
-| Environment Variable | Description                                                                                                                                                    |
-| -------------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------- |
-| `DT_PAAS_TOKEN`      | The token for integrating your Dynatrace environment with Cloud Foundry. You can find it in the deploy Dynatrace section within your environment.              |
-| `DT_SAAS_URL`        | Monitoring endpoint url of the Dynatrace service                                                                                                               |
-| `DT_TENANT`          | Your Dynatrace environment ID is the unique identifier of your Dynatrace environment. You can find it in the deploy Dynatrace section within your environment. |
+| Environment Variable | Description                                                                                   |
+| -------------------- | ----------------------------------------------------------------------------------------------|
+| `DT_PAAS_TOKEN`      | The token for integrating your Dynatrace environment with your Mendix app                     |
+| `DT_SAAS_URL`        | Monitoring endpoint url of the Dynatrace service                                              |
 
-By setting these environment variables automatically the Dynatrace OneAgent will be loaded in the container. 
-
-OneAgent will be able to measure all J2EE related Metrics from the Application. See OneAgent documention for more details. 
+By setting these environment variables, ingestion of metrics to Dynatrace will start.
 
 ### Logging
 The buildpack provides several options to configure logging.
--- a/buildpack/telemetry/dynatrace.py
+++ b/buildpack/telemetry/dynatrace.py
@@ -1,110 +1,25 @@
+"""
+For Dynatrace, metrics are directly ingested through telegraf.
+No additional setup is needed.
+This module only collects information for telegraf from environment variables.
+"""
 import logging
 import os
-import shutil
-import json
+from urllib.parse import urljoin
 
-from buildpack import util
-
-default_env = {
-    "DT_TENANTTOKEN": None,  # optional, default to one found in manifest.json
-    "DT_APPLICATIONID": None,  # optional, default not set
-    "DT_TENANT": None,  # required, environment ID (uuid format)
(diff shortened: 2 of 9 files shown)


etc/telegraf/telegraf.toml.j2
```diff
@@ -297,6 +297,31 @@
...
+{% if dynatrace_enabled %}
+[[outputs.dynatrace]]
+# For usage with the Dynatrace OneAgent you can omit any configuration,
+# the only requirement is that the OneAgent is running on the same host.
+# Only setup environment url and token if you want to monitor a Host without the OneAgent present.
+#
+# Your Dynatrace environment URL.
+# For Dynatrace OneAgent you can leave this empty or set it to "http://127.0.0.1:14499/metrics/ingest" (default)
+# For Dynatrace SaaS environments the URL scheme is "https://{your-environment-id}.live.dynatrace.com/api/v2/metrics/ingest"
+# For Dynatrace Managed environments the URL scheme is "https://{your-domain}/e/{your-environment-id}/api/v2/metrics/ingest"
```
pappachino, line 309: Please clean up the comments here and keep only the ones applicable for SaaS and Managed environments

etc/telegraf/telegraf.toml.j2
```diff
@@ -297,6 +297,31 @@
...
+    url = "{{ dynatrace_ingest_url }}"
+#
+# Your Dynatrace API token.
+# Create an API token within your Dynatrace environment, by navigating to Settings > Integration > Dynatrace API
+# The API token needs data ingest scope permission. When using OneAgent, no API token is required.
+    api_token = "{{ dynatrace_token }}"
+#
+# Optional prefix for metric names (e.g.: "telegraf.")
+# prefix = "telegraf."
+#
```
pappachino, line 319: Please remove if we're not using this.

buildpack/telemetry/dynatrace.py
```diff
@@ -1,110 +1,31 @@
...
-        )
-
-    util.upsert_javaopts(
-        m2ee,
-        [
-            "-agentpath:{path}".format(path=os.path.abspath(agent_path)),
-            "-Xshare:off",
-        ],
-    )
+        logging.debug("Dynatrace is not enabled, cannot find DT_PAAS_TOKEN")
```
pappachino, line 20: This is a bit misleading as `is_enabled()` checks for two parameters. Also we are fetching from environment vars in two separate places. Would be nice to consolidate the vars and checks.

dependencies.yml
```diff
@@ -24,10 +24,6 @@ dependencies:
     trace-agent:
       artifact: datadog/dd-java-agent-{{ version }}.jar
       version: 0.101.0
-  dynatrace:
-    agent:
-      artifact: "{{ url }}/e/{{ environment }}/api/v1/deployment/installer/agent/unix/paas/latest?include=java&bitness=64&Api-Token={{ token }}"
-      managed: false
```
pappachino, line 30: As we're removing the old behavior entirely, it will break for the existing DynaTrace customers. Should we add some backwards compatibility ?

ekremsekerci: I won't expect integration to be broken for existing customers since the same environment variables would be enough for the new approach as well. The names of metrics and the scope can change though. Still, I'd vote to remove old behavior because:

* It's not supported anymore
* It may cause more confusion in the future if we have 2 different ways of integration
* We were not supporting Dynatrace officially

etc/telegraf/telegraf.toml.j2
ekremsekerci: Cleaned up to only keep necessary information

buildpack/telemetry/dynatrace.py
ekremsekerci: Consolidated fetching of two parameters. Added info log to state Dynatrace integration is configured. Still keeping `is_enabled()` separate because it has multiple usages and wanna keep it without logging.

buildpack/telemetry/dynatrace.py
```diff
@@ -1,110 +1,31 @@
...
-    technologies = manifest.get("technologies")
-    java_binaries = technologies.get("java").get("linux-x86-64")
-    for f in java_binaries:
-        binary_type = f.get("binarytype")
-        if binary_type == "loader":
-            return f.get("path")
 
+def ingest_url():
+    if not is_enabled():
+        logging.debug("Dynatrace is not enabled, cannot find DT_SAAS_URL")
```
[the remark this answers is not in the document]
ekremsekerci: Added an info log to state Dynatrace integration is configured

dependencies.yml
pappachino: Sounds good to me. Just need to keep support in the loop.

pappachino (commented):
Looks good to me!

README.md
```diff
@@ -724,15 +724,19 @@ The most important metrics ( `before_xid_wraparound` , `connections` , `database
```
udh-mendix, line 724: Can you remove the block above? Sounds weird IMO.

README.md
```diff
@@ -724,15 +724,19 @@ The most important metrics ( `before_xid_wraparound` , `connections` , `database
...
-By setting these environment variables automatically the Dynatrace OneAgent will be loaded in the container. 
+By setting these environment variables, ingestion of metrics to Dynatrace will start.
 
-OneAgent will be able to measure all J2EE related Metrics from the Application. See OneAgent documention for more details. 
+Buildpack attaches these default tags to metrics that are pushed to Dynatrace:
+
+* app - Environment Id of the Mendix application
+* instance_index - Instance index that metrics belong to
+
+Extra tags can be attached via `TAGS` environment variable (example value: `[env:accp]`)
```
udh-mendix, line 739: Can we add a line for minimum mendix version supported? We have something similar in both datadog and appdynamics docs

README.md
```diff
@@ -724,15 +724,19 @@ The most important metrics ( `before_xid_wraparound` , `connections` , `database
...
+| `DT_SAAS_URL`        | Monitoring endpoint url of the Dynatrace service                                              |
 
-By setting these environment variables automatically the Dynatrace OneAgent will be loaded in the container. 
+By setting these environment variables, ingestion of metrics to Dynatrace will start.
 
-OneAgent will be able to measure all J2EE related Metrics from the Application. See OneAgent documention for more details. 
+Buildpack attaches these default tags to metrics that are pushed to Dynatrace:
+
+* app - Environment Id of the Mendix application
+* instance_index - Instance index that metrics belong to
```
udh-mendix, line 737: make it `app` and `instance_index` for visual clarity

README.md
```diff
@@ -724,15 +724,19 @@ The most important metrics ( `before_xid_wraparound` , `connections` , `database
...
-| `DT_SAAS_URL`        | Monitoring endpoint url of the Dynatrace service                                                                                                               |
-| `DT_TENANT`          | Your Dynatrace environment ID is the unique identifier of your Dynatrace environment. You can find it in the deploy Dynatrace section within your environment. |
+| Environment Variable | Description                                                                                   |
+| -------------------- | ----------------------------------------------------------------------------------------------|
+| `DT_PAAS_TOKEN`      | The token for integrating your Dynatrace environment with your Mendix app                     |
+| `DT_SAAS_URL`        | Monitoring endpoint url of the Dynatrace service                                              |
 
-By setting these environment variables automatically the Dynatrace OneAgent will be loaded in the container. 
+By setting these environment variables, ingestion of metrics to Dynatrace will start.
```
udh-mendix, line 732: line can be removed, its redundant because of line 725.

udh-mendix (approved):

PR #594, opened 2022-12-20
Pass through micrometer and metric_type tags

pappachino:
Allow the following tags to be forwarded along with the metrics to the trends backend : 
- `micrometer_metrics`
- `metric_type`

This allows the backend to know the source of the metric if it is originating from the newer runtimes using the micrometer library or the older runtimes.


ekremsekerci:
looks good to me

ekremsekerci (approved):

Issue #592, opened 2022-12-13
Update Certifi Dependency To Resolve CVE


headkm:
When security scanning the buildpack, [CVE-2022-23491](https://www.cve.org/CVERecord?id=CVE-2022-23491) was discovered. This is due to an older version of Certifi being used. Could Certifi be upgraded from 2021.10.8 to at least 2022.12.07 in the next release?

PR #598, opened 2023-01-09
Bump dependencies

jeohist:
Resolves https://github.com/mendix/cf-mendix-buildpack/issues/592, makes Snyk a little happier

@abhishekamte-mx could you please review my modification to the Data Broker Business Events tests? Just a minor change to make the linter happy 😄

### DIFF
--- a/tests/unit/test_databroker_business_events.py
+++ b/tests/unit/test_databroker_business_events.py
@@ -214,7 +214,7 @@ class TestDataBrokerBusinessEvents(unittest.TestCase):
         assert be_config["BusinessEvents.GenerateMetrics"] == "false"
         assert be_config["BusinessEvents.EnableHeartbeat"] == "false"
 
-    def test_business_events_metrics_constants_config_free(self):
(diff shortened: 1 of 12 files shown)


tests/unit/test_databroker_business_events.py
```diff
@@ -214,7 +214,7 @@ def test_business_events_metrics_constants_config_free(self):
         assert be_config["BusinessEvents.GenerateMetrics"] == "false"
         assert be_config["BusinessEvents.EnableHeartbeat"] == "false"
 
-    def test_business_events_metrics_constants_config_free(self):
+    def test_business_events_metrics_constants_config(self):
```
abhishekamte-mx, line 217: LGTM. just curious, why the linter doesn't like `_free`. maybe we rename it to `_sandbox` ?

jeohist: It's because the previous method is named the same, if there's a better name for it I can modify it too

abhishekamte-mx (approved):

PR #604, opened 2023-02-13
Dynatrace OneAgent integration

ekremsekerci:
- Add support for Dynatrace OneAgent
- Bug fix for Dynatrace Managed

### DIFF
--- a/buildpack/telemetry/dynatrace.py
+++ b/buildpack/telemetry/dynatrace.py
@@ -8,7 +8,7 @@ import os
 import json
 from urllib.parse import urljoin
 
-INGEST_ENDPOINT = "/api/v2/metrics/ingest"
+INGEST_ENDPOINT = "api/v2/metrics/ingest"
 
 
 from buildpack import util
@@ -148,5 +148,22 @@ def get_ingestion_info():
 
     logging.info("Metrics ingestion to Dynatrace via telegraf is configured")
     token = os.getenv("DT_PAAS_TOKEN")
-    ingest_url = urljoin(os.getenv("DT_SAAS_URL"), INGEST_ENDPOINT)
+    ingest_url = _get_ingestion_url(os.getenv("DT_SAAS_URL"), INGEST_ENDPOINT)
     return token, ingest_url
+
+
+def _get_ingestion_url(saas_url, endpoint):
+    """
+    Basic url join but purposefully isolated to add some unittests easily.
+    When merging an url and an additional endpoint, python's urljoin method
+    has so many little details. See:
+    https://stackoverflow.com/questions/10893374/python-confusions-with-urljoin
+
+    So, basically we need to make sure that the url ends with '/' and
+    the endpoint does not start with '/'
+    """
+
+    saas_url = f"{saas_url}/"
+    if endpoint.startswith('/'):
+        endpoint = endpoint[1:]
+    return urljoin(saas_url, endpoint)
--- a/buildpack/telemetry/dynatrace.py
+++ b/buildpack/telemetry/dynatrace.py
@@ -1,17 +1,22 @@
 """
-For Dynatrace, metrics are directly ingested through telegraf.
-No additional setup is needed.
-This module only collects information for telegraf from environment variables.
+For Dynatrace, we have two different ingestion methods:
+1. via Dynatrace OneAgent. It's being downloaded and injected to the java
+runtime.
+2. via telegraf. Telegraf ingests custom runtime metrics using Dynatrace
+output plugin
 """
 import logging
 import os
 import json
+from functools import lru_cache
 from urllib.parse import urljoin
 
-INGEST_ENDPOINT = "api/v2/metrics/ingest"
-
-
 from buildpack import util
+
+INGEST_ENDPOINT = "api/v2/metrics/ingest"
+NAMESPACE = "dynatrace"
+BUILD_PATH = os.path.join(".local", NAMESPACE)
+
 # Environment variables for Dynatrace OneAgent
 # Only passed to the agent if set as environment variable
 default_env = {
@@ -21,40 +26,37 @@ default_env = {
     "DT_TENANT": None,  # required for agent integration, dynatrace environment ID
     "DT_TENANTTOKEN": None,  # optional, default value is get from manifest.json which is downloaded along with the agent installer
     # -- Environment variables for orchestration
-    "DT_LOCALTOVIRTUALHOSTNAME": None,  # optional, default not set
-    "DT_APPLICATIONID": None,  # optional, default not set
-    "DT_NODE_ID": None,  # optional, default not set
     "DT_CLUSTER_ID": None,  # optional, default not set
-    "DT_TAGS": None,  # optional tags e.g. MikesStuff easyTravel=Mike
     "DT_CUSTOM_PROP": None,  # optional metadata e.g. Department=Acceptance Stage=Sprint
     # -- Environment variables for troubleshooting
     "DT_LOGSTREAM": "stdout",  # optional
     "DT_LOGLEVELCON": None,  # Use this environment variable to define the console log level. Valid options are: NONE, SEVERE, and INFO.
     "DT_AGENTACTIVE": None,  # Set to true or false to enable or disable OneAgent.
-    # -- Networking environment variables
-    "DT_NETWORK_ZONE": None,  # optional, Specifies to use a network zone. For more information
-    "DT_PROXY": None, # Optional, When using a proxy, use this environment variable to pass proxy credentials.
 }
 
 
-def stage(buildpack_dir, build_path, cache_path):
+def stage(buildpack_dir, root_dir, cache_path):
     """
     Downloads and unzips necessary OneAgent components
     """
     if is_agent_enabled():
         try:
             util.resolve_dependency(
-                "dynatrace.agent",
-                build_path,  # DOT_LOCAL_LOCATION,
+                dependency="dynatrace.agent",
+                destination=os.path.join(root_dir, NAMESPACE),
                 buildpack_dir=buildpack_dir,
                 cache_dir=cache_path,  # CACHE_DIR,
                 unpack=True,
                 overrides={
-                    "url": os.environ.get("DT_SAAS_URL"),
+                    # need to us rstrip because otherwise the download link formed with double slashes and it doesn't work
+                    "url": os.environ.get("DT_SAAS_URL").rstrip("/"),
                     "environment": os.environ.get("DT_TENANT"),
                     "token": os.environ.get("DT_PAAS_TOKEN"),
                 },
-                ignore_cache=True
+                # cache is not working properly, so ignoring for now.
+                # Can be debugged later, a stack trace exists in the PR:
+                # https://github.com/mendix/cf-mendix-buildpack/pull/562
+                ignore_cache=True,
             )
         except Exception as e:
             logging.warning(
@@ -71,6 +73,7 @@ def update_config(m2ee, app_name):
             "Skipping Dynatrace OneAgent setup, required env vars are not set"
         )
         return
+
     logging.info("Enabling Dynatrace OneAgent")
     try:
         manifest = get_manifest()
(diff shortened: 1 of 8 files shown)


buildpack/telemetry/dynatrace.py
```diff
@@ -5,23 +5,165 @@
...
+                build_path,  # DOT_LOCAL_LOCATION,
+                buildpack_dir=buildpack_dir,
+                cache_dir=cache_path,  # CACHE_DIR,
+                unpack=True,
+                overrides={
+                    "url": os.environ.get("DT_SAAS_URL"),
+                    "environment": os.environ.get("DT_TENANT"),
+                    "token": os.environ.get("DT_PAAS_TOKEN"),
+                },
+                ignore_cache=True
```
pappachino, line 57: Can we check if the caching works now? Or please add a comment on why we are disabling this.

buildpack/telemetry/dynatrace.py
```diff
@@ -5,23 +5,165 @@
...
+            util.upsert_custom_environment_variable(m2ee, key, value)
+    util.upsert_custom_environment_variable(
+        m2ee, "DT_CONNECTION_POINT", get_connection_endpoint()
+    )
+
+    agent_path = os.path.join(".local", get_agent_path())
+    if not os.path.exists(agent_path):
+        raise Exception(
+            "Dynatrace Agent not found: {agent_path}".format(agent_path)
+        )
```
pappachino, line 98: This check should probably go earlier, before we add any custom environment variables or get the manifest file 🤔

tests/unit/test_dynatrace.py
```diff
@@ -5,20 +5,66 @@
...
-        url = "DUMMY_URL"
-        expected_ingest_url = urljoin(url, dynatrace.INGEST_ENDPOINT)
-        env_vars = {"DT_PAAS_TOKEN": token, "DT_SAAS_URL": url}
-        with patch.dict(os.environ, env_vars):
-            actual_info = dynatrace.get_ingestion_info()
+        self.assertFalse(dynatrace.is_telegraf_enabled())
 
-        expected_info = (token, expected_ingest_url)
-        self.assertEqual(expected_info, actual_info)
+    def test_get_ingestion_url(self):
```
pappachino, line 16: Awesome! Nice to have these set of tests. 💯 

For the second URL, we could just refer to the constant in the file. So even if it is modified it still gets validated.

buildpack/telemetry/dynatrace.py
```diff
@@ -5,23 +5,165 @@
...
+    has so many little details. See:
+    https://stackoverflow.com/questions/10893374/python-confusions-with-urljoin
+
+    So, basically we need to make sure that the url ends with '/' and
+    the endpoint does not start with '/'
+    """
+
+    saas_url = f"{saas_url}/"
+    if endpoint.startswith('/'):
+        endpoint = endpoint[1:]
```
pappachino, line 168: This endpoint is a constant and hardcoded, so isn't this check a bit redundant? 
For future scenarios, the unit tests should flag any issues.

And why not refer to the constant directly 🤔 It also simplifies the unit test.

buildpack/telemetry/dynatrace.py
```diff
@@ -5,23 +5,148 @@
...
+    "DT_TAGS": None,  # optional tags e.g. MikesStuff easyTravel=Mike
+    "DT_CUSTOM_PROP": None,  # optional metadata e.g. Department=Acceptance Stage=Sprint
+    # -- Environment variables for troubleshooting
+    "DT_LOGSTREAM": "stdout",  # optional
+    "DT_LOGLEVELCON": None,  # Use this environment variable to define the console log level. Valid options are: NONE, SEVERE, and INFO.
+    "DT_AGENTACTIVE": None,  # Set to true or false to enable or disable OneAgent.
+    # -- Networking environment variables
+    "DT_NETWORK_ZONE": None,  # optional, Specifies to use a network zone. For more information
+    "DT_PROXY": None, # Optional, When using a proxy, use this environment variable to pass proxy credentials.
+}
```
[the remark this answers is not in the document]
ekremsekerci: We are not exposing these directly. In the update_config methods, these are injected into the java process only if they're set as environment variables.

One other point, I couldn't find any proper document on these environment variables. They're mentioned on different pages basically. Not sure all of them are directly valid for our use. Need to test this aspect.

buildpack/telemetry/dynatrace.py
ekremsekerci: I wanted to isolate the behavior regardless of the value of the constant. Basically, if I use the constant here, I won't be able to test its different values.

buildpack/telemetry/dynatrace.py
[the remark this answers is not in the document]
pappachino: If we are not sure if the vars are valid for the scenario, let's not add them to the java process, else it may introduce some unexpected behavior.

Let's keep the env vars to the minimal required ones(we can confirm with the Proximux envvids which ones were being used)

buildpack/telemetry/dynatrace.py
```diff
@@ -5,23 +5,148 @@
...
+            "Failed to parse Dynatrace manifest file", exc_info=True
+        )
+        return
+
+    # dynamic default
+    default_env.update({"DT_TENANTTOKEN": manifest.get("tenantToken")})
+
+    for key, dv in default_env.items():
+        value = os.environ.get(key, dv)
+        if value:
```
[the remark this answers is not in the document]
ekremsekerci: I think we cannot do anything about empty string. If sb sets an environment variable to an empty string explicitly it is a valid input I'd say. 
Boolean values are more ambiguous to me in this case because we get these values from environment variables and inject them into the running java process. I haven't looked for how boolean values can be transferred without being converted to a string but in our case, we already don't have any boolean-valued environment variables.
So, treating any string input from the environment variables as valid seems ok to me in this case.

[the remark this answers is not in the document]
pappachino: Exactly, but this check won't allow empty strings or False values.

So the check should be :
```if value is not None:```

File: buildpack/telemetry/dynatrace.py [hunk 7]
```diff
@@ -5,23 +5,148 @@
...
 import logging
 import os
+import json
 from urllib.parse import urljoin
 
 INGEST_ENDPOINT = "/api/v2/metrics/ingest"
 
 
-def is_enabled():
+from buildpack import util
```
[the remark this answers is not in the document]
ekremsekerci: make format did not fixed this but manually moved to top of constants

buildpack/telemetry/dynatrace.py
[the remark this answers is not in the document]
ekremsekerci: Yeah, that's true. updated

buildpack/telemetry/dynatrace.py
```diff
@@ -5,23 +5,148 @@
...
+    util.upsert_javaopts(
+        m2ee,
+        [
+            "-agentpath:{path}".format(path=os.path.abspath(agent_path)),
+            "-Xshare:off",
+        ],
+    )
+
+
+def get_manifest():
```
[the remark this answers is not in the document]
ekremsekerci: - added cache with 3 as maxsize, just not to think about invalidation, probably not needed but no harm.
- updated path to be based on installation path of dynatrace
- not in favor of adding key-value fetch here in this method since the json has nested dictionaries and every user of this method takes care of their need

buildpack/telemetry/dynatrace.py
```diff
@@ -5,23 +5,148 @@
...
+
+def stage(buildpack_dir, build_path, cache_path):
+    """
+    Downloads and unzips necessary OneAgent components
+    """
+    if is_agent_enabled():
+        try:
+            util.resolve_dependency(
+                "dynatrace.agent",
+                build_path,  # DOT_LOCAL_LOCATION,
```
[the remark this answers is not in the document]
ekremsekerci: added a global variable as `installation_path`. 
Global variable reduces readability a bit since it's randomly assigned in one of the functions. 
However, since we are getting ".local" path from `stage.py` it needs to be set during staging and should be used  in the other methods.
When I checked usage of global variable in my local with a dummy python script it is working like this. Still need to test for the integration since not sure how imports etc affect global variables. Just wanna make sure that it works without any problem.

buildpack/telemetry/dynatrace.py
```diff
@@ -5,23 +5,148 @@
 """
```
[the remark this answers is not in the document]
ekremsekerci: updated

buildpack/telemetry/dynatrace.py
ekremsekerci: added a comment for that. This needs to be investigated separately.

buildpack/telemetry/dynatrace.py
ekremsekerci: moved above

buildpack/telemetry/dynatrace.py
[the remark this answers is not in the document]
pappachino: `functools.lru_cache` is a good approach, but let's use the default max size, else when we add new calls it will have the same issue.

Just so you know `lru_cache` doesn't cache if there is an exception, so all the calls would try to open the file and fail. So won't help in those scenarios.

tests/unit/test_dynatrace.py
pappachino: Right, thats a good point.

buildpack/telemetry/dynatrace.py
```diff
@@ -1,27 +1,181 @@
...
+        0, _get_url(os.environ.get("DT_SAAS_URL"), "communication")
+    )
+    return ";".join(endpoints)
+
+
+def get_agent_path():
+    manifest = get_manifest()
+    technologies = manifest.get("technologies")
+    java_binaries = technologies.get("java").get("linux-x86-64")
+    for f in java_binaries:
```
pappachino, line 140: :nit `f`? Let's have descriptive vars

buildpack/telemetry/dynatrace.py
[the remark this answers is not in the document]
pappachino: I think we've not added `isort` to `make format`. It's ok. This is good now.

buildpack/telemetry/dynatrace.py
```diff
@@ -1,27 +1,181 @@
...
+}
+
+
+def stage(buildpack_dir, build_path, cache_path):
+    """
+    Downloads and unzips necessary OneAgent components
+    """
+    if is_agent_enabled():
+        global installation_path
+        installation_path = os.path.join(build_path, NAMESPACE)
```
pappachino, line 50: Don't you have to define the var at the top once, before overriding the global var 🤔 And It helps in visibility as a top level var.

buildpack/telemetry/dynatrace.py
```diff
@@ -1,27 +1,181 @@
...
 
-    logging.info("Metrics ingestion to Dynatrace is configured")
+    logging.info("Metrics ingestion to Dynatrace via telegraf is configured")
     token = os.getenv("DT_PAAS_TOKEN")
-    ingest_url = urljoin(os.getenv("DT_SAAS_URL"), INGEST_ENDPOINT)
+    ingest_url = _get_url(os.getenv("DT_SAAS_URL"), INGEST_ENDPOINT)
     return token, ingest_url
+
+
+def _get_url(saas_url, endpoint):
```
pappachino, line 167: :nit `_join_url` maybe ?

buildpack/telemetry/dynatrace.py
pappachino: True, injecting vars makes the tests easier. So maybe have the param but default to the top-level constant and do not have the check as the unittests should flag any issues right? Would that work?

buildpack/telemetry/dynatrace.py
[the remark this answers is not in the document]
ekremsekerci: changed

buildpack/telemetry/dynatrace.py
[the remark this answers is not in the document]
ekremsekerci: Only kept the ones used before

buildpack/telemetry/dynatrace.py
ekremsekerci: Changed to lstrip, still behaviour is the same. Just wanna be sure that this method is future proof. If constant changes to have a "/" in the beginning don't affect anything when it's like this

buildpack/telemetry/dynatrace.py
ekremsekerci: changed to file

buildpack/telemetry/dynatrace.py
ekremsekerci: had to abandon global var approach because path during staging was different than the path during runtime

buildpack/telemetry/dynatrace.py
ekremsekerci: updated

pappachino (approved):

PR #613, opened 2023-03-26
WIP: Upgrade to Python 3.10

jeohist:
* Upgrade to Python 3.10
  * Use latest version of pip and pip-tools
    Allows us to remove `_PIP_LOCATIONS_NO_WARN_ON_MISMATCH`
* Replace nosetests with pytest nosetests no longer works under Python 3.10
  pytest is an excellent substitute
* Replace black with ruff ruff has better linting coverage than black and is also a lot faster
* Use Ubuntu 22.04 in builds
  This is to prepare us for the upgrade to cflinuxfs4


jpastoor:
Hi @jeohist,
Really nice how you replaced nosetests with pytest. If we can get this merged I'd like to see how well my mr https://github.com/mendix/cf-mendix-buildpack/pull/612 runs on it. However, can we aim for 3.10.6 instead? That aligns with the runtime installed on the cflinuxfs4 base image.

PR #624, opened 2023-04-19
Route telegraf logs to a file instead of STDERR

pappachino:
`Telegraf` is part of the application metrics pipeline and this change is mainly to suppress telegraf-related warnings and errors from customer logs. 

However, as we still need them for troubleshooting purposes, we route it to a file on disk with logrotate enabled.


udh-mendix (approved):
LGTM

PR #626, opened 2023-04-20
Route `mx_runtime_user_login` metric to trends stack via a separate stream

pappachino:
This PR is more of an optimization of how Mendix collects application runtime metrics for Mendix Cloud graphs.

The change adds a separate telegraf output plugin just for the runtime login metrics as the volume
of these metrics could be quite large and result in metric buffer overflow.


ekremsekerci (approved):
looks good to me

msvolenski (approved):

PR #627, opened 2023-04-25
Add a check to confirm that DEPENDENCY_FILE exists before opening it

Christophe-Vico:
The Buildpack will fail when the file referred to by DEPENDENCY_FILE (while non-compliant content is handled and will return an empty).
This PR adds a check of existence of DEPENDENCY_FILE. If the files does not exists, an empty list is returned


Christophe-Vico:
This does not fix the problem further in mx_java_agent.py / line 69, when utils.py/get_dependency returns None

PR #629, opened 2023-05-03
APM metrics filters

ekremsekerci:
Add filtering capability for metrics ingestion to third-party APMs

APM_METRICS_FILTER_ALLOW and APM_METRICS_FILTER_DENY environment variables are introduced.

Also in telegraf configuration round_interval set to true for internal metrics

### DIFF
--- a/README.md
+++ b/README.md
@@ -583,7 +583,7 @@ Note that:
 
 ## Telemetry Configuration
 
-The buildpack includes a variety of telemetry agents, and can configure logging for the Mendix Runtime.
+The buildpack includes a variety of telemetry agents that can be configured to collect and forward metrics/logs from the Mendix Runtime.
 
 ### New Relic
 
@@ -828,6 +828,43 @@ Example (1000 loglines/second):
 ```shell
 cf set-env <YOUR_APP> LOG_RATELIMIT '1000'
 ```
+
+### Custom Runtime Metrics filtering
+
+For the third-party integrations explained above, in addition to the metrics collected by the agents, custom runtime metrics are provided via telegraf.
+This configuration also has a filtering mechanism that allows users to specify metrics they allow or deny for the vendor they are using.
+To filter the ingestion of custom runtime metrics to third party APMs, users should provide a list of prefixes of the metrics they want to allow/deny using the environment variables listed below.
+
+Note: Custom database metrics cannot be filtered by name, to turn them off, the `APPMETRICS_INCLUDE_DB` environment variable should be set to false. 
+
+#### APM_METRICS_FILTER_ALLOW
+
+Comma-separated list of prefixes for the metrics to be allowed. By default, all metrics are allowed, even if they are not specified via this env var.
+
+For example, to allow only the session metrics, `APM_METRICS_FILTER_ALLOW` should be set to `mx.runtime.stats.sessions`:
+
+```shell
+cf set-env <YOUR_APP> APM_METRICS_FILTER_ALLOW 'mx.runtime.stats.sessions'
+```
+
+#### APM_METRICS_FILTER_DENY
+
+Comma-separated list of prefixes for the metrics to be denied. 
+
+For example, to deny all metrics starting with jetty or mx.runtime, the environment variable should be set to `jetty,mx.runtime`:
+
+```shell
+cf set-env <YOUR_APP> APM_METRICS_FILTER_DENY 'jetty,mx.runtime'
+```
+
+#### APM_METRICS_FILTER_DENY_ALL
+
+If this environment variable is set to `true`, all metrics will be denied regardless of values of `APM_METRICS_FILTER_ALLOW`, `APM_METRICS_FILTER_DENY`, and `APPMETRICS_INCLUDE_DB`.
+
+```shell
+cf set-env <YOUR_APP> APM_METRICS_FILTER_DENY_ALL true
+```
+
 
 ## Using the Buildpack without an Internet Connection
 
--- a/buildpack/telemetry/metrics.py
+++ b/buildpack/telemetry/metrics.py
@@ -20,69 +20,6 @@ from lib.m2ee.util import strtobool
 
 from . import datadog, appdynamics, dynatrace
 
-# Runtime configuration for influx registry
-# This enables the new stream of metrics coming from micrometer instead
-# of the admin port.
-# https://docs.mendix.com/refguide/metrics#registries-configuration
-# NOTE: Metrics are usually dot separated. But each registry has its
-# own naming format. For instance, a metric like
-# `a.name.like.this` would appear as `a_name_like_this` in
-# influx-formatted metrics output. Hence the filter names uses the
-# dot-separated metric names.
-INFLUX_REGISTRY = {
-    "type": "influx",
-    "settings": {
-        "uri": "http://localhost:8086",
-        "db": "mendix",
-        "step": "10s",
-    },
-    "filters": [
-        # Login metrics needs to be enabled explicitly as it's disabled
-        # by default
-        {
-            "type": "nameStartsWith",
-            "result": "accept",
-            "values": ["mx.runtime.user.login"],
-        },
-        # Filter out irrelevant metrics to reduce
-        # the payload size passed to TSS/TFR
-        # https://docs.mendix.com/refguide/metrics#filters
-        {
-            "type": "nameStartsWith",
-            "result": "deny",
-            "values": ["commons.pool", "jvm.buffer"],
-        },
-    ],
-}
-
-STATSD_REGISTRY = {
-    "type": "statsd",
-    "settings": {"port": datadog.get_statsd_port()},
-}
-
-# For freeapps we push only the session & login metrics
-FREEAPPS_METRICS_REGISTRY = [
-    {
-        "type": "influx",
-        "settings": {
-            "uri": "http://localhost:8086",
-            "db": "mendix",
-            "step": "10s",
-        },
-        "filters": [
-            {
-                "type": "nameStartsWith",
-                "result": "accept",
-                "values": [
-                    "mx.runtime.stats.sessions",
-                    "mx.runtime.user.login",
-                ],
-            },
-            {"type": "nameStartsWith", "result": "deny", "values": [""]},
-        ],
-    }
-]
-
 METRICS_REGISTRIES_KEY = "Metrics.Registries"
 
 # From this MxRuntime version onwards we gather (available) runtime statistics
@@ -186,9 +123,9 @@ def configure_metrics_registry(m2ee):
 
     logging.info("Configuring runtime to push metrics to influx via micrometer")
     if util.is_free_app():
-        return FREEAPPS_METRICS_REGISTRY
+        return [get_freeapps_registry()]
 
-    paidapps_registries = [INFLUX_REGISTRY]
+    paidapps_registries = [get_influx_registry()]
 
     if (
         datadog.is_enabled()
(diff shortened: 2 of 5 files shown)


buildpack/telemetry/metrics.py
```diff
@@ -55,9 +55,37 @@
...
+    it accepts/denies every metric since we us type as `nameStartsWith`.
+    To prevent breaking the functionality because of this, we need to make sure
+    that we pass empty string to the registry filters only if it's intentional.
+    So, we strip the leading and trailing commas. Additionally we remove all
+     the white spaces to prevent any unintentional mistakes.
+    """
+    env_var_value = os.getenv(env_var_name)
+    if env_var_value is None:
+        return []
+    return env_var_value.replace(" ", "").strip(",").split(",")
```
pappachino, line 71: Not sure why we need the additional `strip` 🤔 

This should work right ?
```suggestion
    return env_var_value.replace(" ", "").split(",")
```

pappachino (changes requested):
Can we add some unit tests similar to what we have [here](https://github.com/mendix/cf-mendix-buildpack/blob/master/tests/unit/test_micrometer_metrics.py#L65)

buildpack/telemetry/metrics.py
ekremsekerci: If the there is a leading/trailing comma then split creates empty string in the resulting list which breaks our logic since the empty string means all metrics in this case. 
image

buildpack/telemetry/metrics.py
```diff
@@ -55,9 +55,37 @@
     ],
 }
 
+
+def get_apm_metrics_filter(env_var_name):
+    """
+    If we use empty string ("") in the filters that we use for statsd registry,
+    it accepts/denies every metric since we us type as `nameStartsWith`.
```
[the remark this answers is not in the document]
ekremsekerci: Either I'll add a similar explanation to the docs or refer to this.

Fixed the typo, thanks

buildpack/telemetry/metrics.py
```diff
@@ -55,9 +55,37 @@
...
+            "type": "nameStartsWith",
+            "result": "accept",
+            "values": get_apm_metrics_filter("APM_METRICS_FILTER_ALLOW"),
+        },
+        {
+            "type": "nameStartsWith",
+            "result": "deny",
+            "values": get_apm_metrics_filter("APM_METRICS_FILTER_DENY"),
+        },
+    ],
```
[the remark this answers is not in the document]
ekremsekerci: Yeah, then it's not effective. Will add this case to the peer testing steps, thanks 👍

ekremsekerci:
> Can we add some unit tests similar to what we have [here](https://github.com/mendix/cf-mendix-buildpack/blob/master/tests/unit/test_micrometer_metrics.py#L65)

The unit tests we have there are to determine which registries will be used under which conditions. With my changes, I don't create any other possibility. I'm just playing with the existing statsd registry.

Some unit tests can be added to test the method `get_apm_metrics_filter`. However, those tests mostly will be like testing `replace`, `strip`, and `split` functionalities so I was not so sure if adding those tests add much value. If you think it adds value I can add?

buildpack/telemetry/metrics.py
pappachino: Ah okay, got it. It's a bit of over-optimization though as to why we only care about the comma 😄

pappachino:
> > Can we add some unit tests similar to what we have [here](https://github.com/mendix/cf-mendix-buildpack/blob/master/tests/unit/test_micrometer_metrics.py#L65)
> 
> The unit tests we have there are to determine which registries will be used under which conditions. With my changes, I don't create any other possibility. I'm just playing with the existing statsd registry.
> 
> Some unit tests can be added to test the method `get_apm_metrics_filter`. However, those tests mostly will be like testing `replace`, `strip`, and `split` functionalities so I was not so sure if adding those tests add much value. If you think it adds value I can add?

Yeah, mainly to ensure we are applying the expected filters when we set the ALLOW and DENY env vars. Better to add some unit-tests to avoid any future regressions.

buildpack/telemetry/metrics.py
```diff
@@ -55,9 +55,37 @@
...
+    env_var_value = os.getenv(env_var_name)
+    if env_var_value is None:
+        return []
+    return env_var_value.replace(" ", "").strip(",").split(",")
+
+
 STATSD_REGISTRY = {
     "type": "statsd",
     "settings": {"port": datadog.get_statsd_port()},
+    "filters": [
```
pappachino, line 77: Can we log the final filter values only if the user has provided any of the envvars? Would help us debug if the user has entered unexpected patterns.

pappachino (changes requested):

buildpack/telemetry/metrics.py
ekremsekerci: We only care about comma because we split by comma :) We'll be waiting for comma separated values as input and the logic breaks if there is a leading/trailing comma in the input. So basically, I'm trying to prevent unintentional errors

buildpack/telemetry/metrics.py
pappachino: I think `INFO` is fine too, as it should show up only once during startup right? Else we need to ask them to enable `DEBUG` to figure out any issue.

etc/telegraf/telegraf.toml.j2
```diff
@@ -29,6 +29,7 @@
     commands = ["echo {{ is_enabled }}"]
     name_override = "integration_usage"
     interval = "6h"
+    round_interval = false
```
pappachino, line 32: A note on why we overriding this would be helpful

README.md
```diff
@@ -820,6 +820,30 @@ Example (1000 loglines/second):
 cf set-env <YOUR_APP> LOG_RATELIMIT '1000'
 ```
 
+### Custom Runtime Metrics filtering
+
+For the third-party integrations explained above, in addition to the metrics collected by the agents, custom runtime metrics are provided via telegraf.
+This configuration also has a filtering mechanism that allows users to specify metrics they allow or deny for the vendor they are using.
+To be able to filter the emission of custom runtime metrics, users should provide list of prefixes of metrics they need to allow/deny using the environment variables listed below.
```
msvolenski, line 827: "To filter the emission of custom runtime metrics, users should provide a list of prefixes of the metrics they want to allow/deny using the environment variables listed below."

Just a minor rewriting.

README.md
```diff
@@ -820,6 +820,30 @@ Example (1000 loglines/second):
...
+
+Note: Custom database metrics cannot be filtered by name, to turn them off `APPMETRICS_INCLUDE_DB` environment variable should be set to false.
```
msvolenski, line 829: "Note: Custom database metrics cannot be filtered by name, to turn them off, the `APPMETRICS_INCLUDE_DB` environment variable should be set to false."

Just a minor rewriting.

README.md
```diff
@@ -820,6 +820,30 @@ Example (1000 loglines/second):
...
+This configuration also has a filtering mechanism that allows users to specify metrics they allow or deny for the vendor they are using.
+To be able to filter the emission of custom runtime metrics, users should provide list of prefixes of metrics they need to allow/deny using the environment variables listed below.
+
+Note: Custom database metrics cannot be filtered by name, to turn them off `APPMETRICS_INCLUDE_DB` environment variable should be set to false. 
+
+#### APM_METRICS_FILTER_DENY
+
+List of prefixes for the metrics to be denied. 
+
+For example to deny all metrics starting with **jetty** or **mx.runtime** the env var should be set to `jetty,mx.runtime`.
```
msvolenski, line 835: "For example, to deny all metrics starting with **jetty** or **mx.runtime**, the environment variable should be set to `jetty,mx.runtime`."

Just a minor rewriting.

README.md
```diff
@@ -820,6 +820,30 @@ Example (1000 loglines/second):
...
+
+To deny all the custom metrics, the environment variable should be set without any value.
+
+
+
+#### APM_METRICS_FILTER_ALLOW
+
+List of prefixes for the metrics to be allowed. By default, all metrics are allowed even if they are not specified via this env var, so it makes sense to use this environment variable in combination with the `APM_METRICS_FILTER_DENY`.
```
msvolenski, line 843: "List of prefixes for the metrics to be allowed. By default, all metrics are allowed, even if they are not specified via this env var, so it makes sense to use this environment variable in combination with the `APM_METRICS_FILTER_DENY`."

Just a minor rewriting.

README.md
```diff
@@ -820,6 +820,30 @@ Example (1000 loglines/second):
...
+
+For example to deny all metrics except the session metrics, `APM_METRICS_FILTER_DENY` should be used without any value as explained before and `APM_METRICS_FILTER_ALLOW` should be set to `mx.runtime.stats.sessions`.
```
msvolenski, line 845: "For example, to deny all metrics except the session metrics, `APM_METRICS_FILTER_DENY` should be used without any value as explained before, and `APM_METRICS_FILTER_ALLOW` should be set to `mx.runtime.stats.sessions`."

Just a minor rewriting.

buildpack/telemetry/metrics.py
```diff
@@ -186,21 +123,119 @@ def configure_metrics_registry(m2ee):
...
+    if env_var_value is None:
+        return []
+    apm_filter = env_var_value.replace(" ", "").strip(",").split(",")
+    logging.info("APM metrics filter for %s is: %s", env_var_value, apm_filter)
+    return apm_filter
+
+
+def get_freeapps_registry():
+    # For freeapps we push only the session & login metrics
+    return [
```
msvolenski, line 216: I know you didn't change that, but I would like to suggest that we return a dictionary here and add the list in the method that calls it, as it's done for the others.

tests/unit/test_micrometer_metrics.py
```diff
@@ -97,12 +97,43 @@ def test_paidapps_metrics_registry_statsd(self, is_enabled):
...
+            {
+                "PROFILE": "some-random-mx-profile",
+                "APM_METRICS_FILTER_ALLOW": "allowed_metric",
+                "APM_METRICS_FILTER_DENY": "denied_metric",
+            },
+        ):
+            result = metrics.configure_metrics_registry(Mock())
+            metrics_registries = sorted(result, key=itemgetter("type"))
+
+            # influx registry shouldn't the filters
```
msvolenski, line 113: "[...] shouldn't have [...]"

etc/telegraf/telegraf.toml.j2
ekremsekerci: removed this already

README.md
```diff
@@ -820,6 +820,30 @@ Example (1000 loglines/second):
...
+
+Note: Custom database metrics cannot be filtered by name, to turn them off `APPMETRICS_INCLUDE_DB` environment variable should be set to false. 
+
+#### APM_METRICS_FILTER_DENY
+
+List of prefixes for the metrics to be denied. 
+
+For example to deny all metrics starting with **jetty** or **mx.runtime** the env var should be set to `jetty,mx.runtime`.
+
+To deny all the custom metrics, the environment variable should be set without any value.
```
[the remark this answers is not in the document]
ekremsekerci: added

README.md
```diff
@@ -820,6 +820,30 @@ Example (1000 loglines/second):
...
+
+
+
+#### APM_METRICS_FILTER_ALLOW
```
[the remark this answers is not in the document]
ekremsekerci: added

README.md
ekremsekerci: updated

README.md
ekremsekerci: updated

README.md
ekremsekerci: updated

README.md
ekremsekerci: updated

README.md
ekremsekerci: updated

buildpack/telemetry/metrics.py
ekremsekerci: updated, thanks for the suggestion 👍

tests/unit/test_micrometer_metrics.py
ekremsekerci: updated

README.md
```diff
@@ -824,36 +824,31 @@ cf set-env <YOUR_APP> LOG_RATELIMIT '1000'
 
 For the third-party integrations explained above, in addition to the metrics collected by the agents, custom runtime metrics are provided via telegraf.
 This configuration also has a filtering mechanism that allows users to specify metrics they allow or deny for the vendor they are using.
-To filter the emission of custom runtime metrics, users should provide a list of prefixes of the metrics they want to allow/deny using the environment variables listed below.
+To filter the ingestion of custom runtime metrics to third party APMs, users should provide a list of prefixes of the metrics they want to allow/deny using the environment variables listed below.
```
pappachino, line 827: 💯 Yeah emission for some reason didn't sound right. This is good.

README.md
```diff
@@ -824,36 +824,31 @@ cf set-env <YOUR_APP> LOG_RATELIMIT '1000'
...
 Note: Custom database metrics cannot be filtered by name, to turn them off, the `APPMETRICS_INCLUDE_DB` environment variable should be set to false. 
 
-#### APM_METRICS_FILTER_DENY
+#### APM_METRICS_FILTER_ALLOW
 
-List of prefixes for the metrics to be denied. 
+List of prefixes for the metrics to be allowed. By default, all metrics are allowed, even if they are not specified via this env var.
 
-For example, to deny all metrics starting with jetty or mx.runtime, the environment variable should be set to `jetty,mx.runtime`:
+For example, to deny all metrics except the session metrics, `APM_METRICS_FILTER_ALLOW` should be set to `mx.runtime.stats.sessions`:
```
pappachino, line 835: As we're in the `ALLOW` section, let's rephrase it to
```suggestion
For example, to allow only the session metrics, `APM_METRICS_FILTER_ALLOW` should be set to `mx.runtime.stats.sessions`:
```

README.md
```diff
@@ -824,36 +824,31 @@ cf set-env <YOUR_APP> LOG_RATELIMIT '1000'
...
-To filter the emission of custom runtime metrics, users should provide a list of prefixes of the metrics they want to allow/deny using the environment variables listed below.
+To filter the ingestion of custom runtime metrics to third party APMs, users should provide a list of prefixes of the metrics they want to allow/deny using the environment variables listed below.
 
 Note: Custom database metrics cannot be filtered by name, to turn them off, the `APPMETRICS_INCLUDE_DB` environment variable should be set to false. 
 
-#### APM_METRICS_FILTER_DENY
+#### APM_METRICS_FILTER_ALLOW
 
-List of prefixes for the metrics to be denied. 
+List of prefixes for the metrics to be allowed. By default, all metrics are allowed, even if they are not specified via this env var.
```
pappachino, line 833: Let's rephrase it to
```suggestion
Comma-separated list of prefixes for the metrics to be allowed. By default, all metrics are allowed, even if they are not specified via this env var.
```

README.md
```diff
@@ -824,36 +824,31 @@ cf set-env <YOUR_APP> LOG_RATELIMIT '1000'
...
-cf set-env <YOUR_APP> APM_METRICS_FILTER_DENY ''
+cf set-env <YOUR_APP> APM_METRICS_FILTER_ALLOW 'mx.runtime.stats.sessions'
 ```
 
+#### APM_METRICS_FILTER_DENY
 
-#### APM_METRICS_FILTER_ALLOW
-
-List of prefixes for the metrics to be allowed. By default, all metrics are allowed, even if they are not specified via this env var, so it makes sense to use this environment variable in combination with the `APM_METRICS_FILTER_DENY`.
+List of prefixes for the metrics to be denied.
```
pappachino, line 843:
```suggestion
Comma-separated list of prefixes for the metrics to be denied. 
```

buildpack/telemetry/metrics.py
```diff
@@ -176,25 +176,41 @@ def get_influx_registry():
 
 
 def get_statsd_registry():
+    allowed_metrics = os.getenv("APM_METRICS_FILTER_ALLOW")
+    denied_metrics = os.getenv("APM_METRICS_FILTER_DENY")
+
+    if allowed_metrics and not denied_metrics:
```
pappachino:
```suggestion
    if allowed_metrics and denied_metrics is not None:
```
More of a nitpick. Do we need to support falsey prefixes like `0` or `false`

buildpack/telemetry/metrics.py
```diff
@@ -176,25 +176,41 @@ def get_influx_registry():
 
 
 def get_statsd_registry():
+    allowed_metrics = os.getenv("APM_METRICS_FILTER_ALLOW")
```
pappachino: Let's keep this allow and deny logic outside of `get_statsd_registry` method, as it is not tied to the registry functionality and can be re-used if needed.

buildpack/telemetry/metrics.py
ekremsekerci: Here we're looking for (denied_metrics is None), but yeah for being more explicit, I'm refactoring my version to this. 

I think we don't need to validate the content here because it can easily be understood by users and fixed even if they do some mistakes on the first attempts

README.md
ekremsekerci: applied

README.md
ekremsekerci: applied

buildpack/telemetry/metrics.py
ekremsekerci: isolated the logic in a separate method now. get_statsd_registry can be called with empty lists if we need to use it anywhere else without filters

README.md
```diff
@@ -848,6 +848,16 @@ For example, to deny all metrics starting with jetty or mx.runtime, the environm
...
 ```
 
+#### APM_METRICS_FILTER_DENY_ALL
+
+If this environment variable is set to `true`, all metrics will be denied regardless of values of `APM_METRICS_FILTER_ALLOW` and `APM_METRICS_FILTER_DENY`. 
+
+Be aware that `APPMETRICS_INCLUDE_DB` still needs to be set to false to turn off the database metrics.
+
+```shell
+cf set-env <YOUR_APP> APM_METRICS_FILTER_DENY 'jetty,mx.runtime'
```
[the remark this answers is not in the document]
ekremsekerci: updated

buildpack/telemetry/telegraf.py
```diff
@@ -75,6 +75,8 @@ def _get_config_file_path(version):
 
 
 def include_db_metrics():
+    if metrics.deny_all_apm_metrics():
```
msvolenski, line 78: This could be put in the same condition below, but I understand keeping the free apps in a separate logic.
So I'm fine with both.

ekremsekerci: yeah, I just wanted to make it very explicit and separate in the beginning since this env var precedes all the other logic

pappachino (approved):

PR #644, opened 2023-06-14
send dependencies.json in client-config request

AlbertDeVries2:
the client-config request to the cf-kafka-broker is updated to a PUT instead of a GET that excepts the dependencies.json file in the body. The start is modified to read this file and send it in the request.
note: shouldn't be merged before the cf-kafka-broker changes are live

### DIFF
--- a/buildpack/databroker/business_events.py
+++ b/buildpack/databroker/business_events.py
@@ -45,7 +45,6 @@ def _put_client_config(url, auth_token, version, dependencies_json_str):
             backoff_factor=0.1,
             status_forcelist=[ 500, 502, 503, 504 ])
     adapter = HTTPAdapter(max_retries=retries)
-    session.mount('http://', adapter)
     session.mount('https://', adapter)
 
     resp = session.put(
--- a/buildpack/databroker/business_events.py
+++ b/buildpack/databroker/business_events.py
@@ -41,19 +41,15 @@ def _put_client_config(url, auth_token, version, dependencies_json_str):
     }
 
     session = requests.Session()
-    retries = Retry(total=2,
-            backoff_factor=0.1,
-            status_forcelist=[ 500, 502, 503, 504 ])
+    retries = Retry(total=2, backoff_factor=0.1, status_forcelist=[500, 502, 503, 504])
     adapter = HTTPAdapter(max_retries=retries)
-    session.mount('https://', adapter)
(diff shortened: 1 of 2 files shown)


buildpack/databroker/business_events.py
```diff
@@ -55,6 +59,17 @@ def _configure_business_events_metrics(be_config, existing_constants):
             be_config[f"{CONSTANTS_PREFIX}.EnableHeartbeat"] = "true"
 
 
+def _read_dependencies_json():
+    file_path = os.path.join(BASE_PATH, "model", "dependencies.json")
+    try:
+        with open(file_path) as f:
+            return json.load(f)
```
abhishekamte-mx, line 66: we need the contents of dependencies json as a string and not a json, as thats how the instance table is maintained with request coming in from update api call. so instead maybe we should make the file contents as a parameter, check the cf-kafka-broker MR review comments :) for my suggestion

buildpack/databroker/business_events.py
```diff
@@ -30,16 +34,16 @@ def update_config(m2ee, vcap_services_data):
     logging.debug("Business Events config added to MicroflowConstants")
 
 
-def _get_client_config(url, auth_token, version):
+def _put_client_config(url, auth_token, dependencies_json):
     headers = {
-        "Authorization": f"Bearer {auth_token}",
-        "X-Version": version,
```
abhishekamte-mx, line 36: any specific reason you feel versioning is not required now?

abhishekamte-mx (changes requested):

buildpack/databroker/business_events.py
AlbertDeVries2: yeah good question, i guess it was me being weak not to know on which version to start so i removed it all together. Version 1 is now for the get, should this be a version 1 PUT? or version 2 PUT? what do you think?

buildpack/databroker/business_events.py
AlbertDeVries2: yes good suggestion, ill re-write it to be string at this level

AlbertDeVries2: it makes no sense to parse it here

sailhenz (approved):
lgtm

sailhenz (approved):

PR #647, opened 2023-06-19
Set garbage collector based on container memory or provided env variable [v4]

sailhenz:


mbyregowMx (approved):
Changes look good, approved

PR #649, opened 2023-06-20
fix: Revert "Set garbage collector based on container memory or provided env variable [v4]"

jpastoor:
Reverts mendix/cf-mendix-buildpack#647
Due to issues in production.


Yornik (approved):
revert

PR #651, opened 2023-06-20
fix: fixed the removal of the skip-duplicate-actions in release

jpastoor:
I made an error in my initial change to bypass `skip-duplicate-actions`.


Yornik (commented):
ok

Yornik (approved):
ok looks ok

PR #652, opened 2023-06-20
restored: set garbage collector based on container memory or provided env variable

jpastoor:
Restored the earlier reverted commit [set garbage collector based on container memory or provided env variable](https://github.com/mendix/cf-mendix-buildpack/commit/f4f276886e20ebfda6979d1bc4e3c8ca86a9415a)


Yornik (approved):
Revert of the Revert

PR #653, opened 2023-06-26
Db 2086 send dependencies json in client config call release 4.x

AlbertDeVries2:
the client-config request to the cf-kafka-broker is updated to a PUT instead of a GET that excepts the dependencies.json file in the body. The start is modified to read this file and send it in the request.
note: shouldn't be merged before the cf-kafka-broker changes are live


sailhenz (dismissed):
lgtm

sailhenz (approved):

PR #664, opened 2023-07-05
send dependencies.json in client-config PUT request

AlbertDeVries2:
same code as before, re-enable the new version of the call


sailhenz (approved):
lgtm

PR #674, opened 2023-09-12
COA2-865: remove unused ca

Yornik:
I removed  the certificates that we don't use right now but added the CA for me-central-1
Original ones:
https://truststore.pki.rds.amazonaws.com/af-south-1/af-south-1-bundle.pem
https://truststore.pki.rds.amazonaws.com/me-south-1/me-south-1-bundle.pem
https://www.amazontrust.com/repository/AmazonRootCA1.pem
New one:
https://truststore.pki.rds.amazonaws.com/me-central-1/me-central-1-bundle.pem

Also, cherry-picked @pappachino commits to fix the integration test.


djvdorp:
I can confirm that this change is the minimum required change to add the `me-central-1` region which means literally include the content from the following file: https://truststore.pki.rds.amazonaws.com/me-central-1/me-central-1-bundle.pem

This can be confirmed by hand, but since we have had a point in time where there were many (most likely all available ones) were added and then subsequently removed, the diff is hard to read, so I decided to compare this PR against the state before the former PR instead (https://github.com/mendix/cf-mendix-buildpack/pull/668) to have the smallest diff possible. That would be be revision `183f4e5af7f47f47367fcc45a3cc8746aa69137f` which is the last change on file `etc/amazon-rds-ca.pem` before `35e30369c59c13a2f888d716709bee0f21bf80c9` (which is in https://github.com/mendix/cf-mendix-buildpack/pull/668). 

The diff matches the content of the [AWS-supplied file](https://truststore.pki.rds.amazonaws.com/me-central-1/me-central-1-bundle.pem) exactly, so that is confirmed watertight, and can be verified via:

```
$ git diff 183f4e5af7f47f47367fcc45a3cc8746aa69137f..7395996bc5c9da79b94756a3b97b27fc6777ae4e etc/amazon-rds-ca.pem
```

mukund-padale (approved):

PR #677, opened 2023-10-02
fix: set the default keepalive to 100s

loadpi:
see https://docs.cloudfoundry.org/adminguide/routing-keepalive.html#app-idle-timeout


sailhenz (approved):
LGTM

PR #685, opened 2023-10-25
New Relic Integration

msvolenski:
Add support to New Relic integration in the Mendix Cloud.
- JVM metrics
- Database metrics
- Custom Mendix metrics
- Logs

Deprecation of `SPLUNK_LOGS_REDACTION` and `DATADOG_LOGS_REDACTION` in favor of `LOGS_REDACTION`. Which is now also applicable to the New Relic integration.


pappachino (approved):
Looks good!!

pappachino (approved):

buildpack/telemetry/datadog.py
```diff
@@ -80,6 +80,15 @@ def _is_tracing_enabled():
 
 # Toggles logs redaction (email addresses are replaced by a generic string)
 def _is_logs_redaction_enabled():
+    """Check if logs should be redacted."""
+
+    # Use this, if it is set
+    logs_redaction = os.getenv("LOGS_REDACTION")
+    if logs_redaction is not None:
+        return strtobool(logs_redaction)
```
ajinkya-takle, line 88: Why no bool() over here, as `strtobool` returns integer? Is it because DD needs true/false value as 1/0?

msvolenski: DD doesn't care, actually. The result of this method is used in an `if` statement, so both true/false and 1/0 would be good.
Since it was already like this, I decided not to change the return type.

PR #699, opened 2023-11-23
add support for the ca certificates on all regions.

Yornik:
Using the pems on : https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/UsingWithRDS.SSL.html

And some logic to select the right PEM file based on the region in the endpoint.
@djvdorp and @azourigal were in a meeting reviewing this.
We are discussing this internally. Please dm us on Slack if there are questions.


buildpack/infrastructure/database.py
```diff
@@ -303,10 +335,11 @@ def init(self):
         if database_type == "PostgreSQL" and config["DatabaseHost"].split(":")[
             0
         ].endswith(".rds.amazonaws.com"):
+            database_region = config["DatabaseHost"].split('.')[2]
             jdbc_params.update(
                 {
                     "sslrootcert": os.path.expandvars(
-                        "$HOME/.postgresql/amazon-rds-ca.pem"
+                        "$HOME/.postgresql/"+region_pem_map[database_region]
```
jeohist, line 342: Why not something like `f"$HOME/.postgresql/{database_region}-bundle.pem"`?

Yornik: We did discuss this because we need to import/write the pem files to the /etc manually to keep the build pack offline. so we wanted to make it more explicit what we are doing.

mxsoldaat (approved):
The proposed changes look good to me

PR #701, opened 2023-11-28
Added Non Mendix Public cloud Flag to set metric type

gauravmendix:
Added flag "NON_MENDIX_PUBLIC_CLOUD" to decide whether micrometer metrics is enabled. This is required because metrics url is not set on SAP BTP so need another check to set it to micrometer for Mendix runtime > 9.7


buildpack/telemetry/metrics.py
```diff
@@ -108,9 +108,18 @@ def _micrometer_runtime_requirement(runtime_version):
...
+    if(_micrometer_runtime_requirement(runtime_version)):
+        logging.info("satisfies micrometer runtime requirement")
+        if(bool(get_micrometer_metrics_url())):
+            logging.info("Found micrometer metrics url configured")
+            micrometer_enabled = True
+        elif(strtobool(os.getenv("NON_MENDIX_PUBLIC_CLOUD","false"))):
+            logging.info("micrometer for non mendix public cloud")
+            micrometer_enabled =  True
+    return micrometer_enabled
+
```
pappachino, line 122: Not sure if we need these many `INFO` log lines here. Can we have it as `DEBUG` instead?

gauravmendix: Changed to debug. Thanks!

pappachino (approved):

PR #705, opened 2023-12-11
admin API should listen on all interfaces

xiwenc:
Admin interface can be exposed safely on all interfaces in a container model because it requires a password to access privileged functions. This is needed for the `/Prometheus` endpoint to work. Because pod scraping by Prometheus is initiated outside of the pod.


jeohist (approved):

jeohist:
Thanks for your contribution @xiwenc ! It will be included in our next release.

PR #711, opened 2024-01-12
Updated the maintained versions to match the documentation

Yornik:
The versions in https://docs.mendix.com/releasenotes/studio-pro/lts-mts did not match the versions we checked for in the build pack. This is corrected.


jeohist:
@Yornik could you update the tests in https://github.com/mendix/cf-mendix-buildpack/blob/develop/tests/unit/test_deprecations.py as well?

Yornik:
> @Yornik could you update the tests in https://github.com/mendix/cf-mendix-buildpack/blob/develop/tests/unit/test_deprecations.py as well?

Will do thank you

tests/unit/test_deprecations.py
```diff
@@ -27,9 +27,6 @@ def test_mx7_supported(self):
 
 
 class TestCaseMxMaintained(TestCase):
-    def test_mx7_maintained(self):
-        assert runtime.is_version_maintained(MXVersion("7.23.1"))
-
```
pommi, line 32: Let's move this one to `test_mx7_not_maintained`

tests/unit/test_deprecations.py
```diff
@@ -40,10 +37,13 @@ def test_mx8_not_maintained(self):
         assert not runtime.is_version_maintained(MXVersion("8.17"))
 
     def test_mx9_maintained(self):
-        assert runtime.is_version_maintained(MXVersion("9.6.9"))
-        assert runtime.is_version_maintained(MXVersion("9.12.1"))
```
pommi, line 44: Could you move these to `test_mx9_not_maintained`?

tests/unit/test_deprecations.py
Yornik: for extended support we will still support this. Do we want customers to see a line that it is unsupported in that case?

jeohist: Should we add an extended support function / test? 🤔

pommi: Good idea @jeohist ! @Yornik Let's do it in another PR, revert this chunk and revert https://github.com/mendix/cf-mendix-buildpack/pull/711/files#diff-54130e27ebbe04d07ebe3741504e8adc20f28fa9cd586a5f6d0ad14584be6878L43-L44 as well.

Yornik: reverted

tests/unit/test_deprecations.py
Yornik: done

pommi (approved):

PR #718, opened 2024-03-18
Pref.xml template Update for Non-Mendix Cloud (SAP)

bhavinshah-mendix:
Added code in runtime file, where it will update license key and id in pref.xml file for non-mendix cloud

### DIFF
--- a/buildpack/core/runtime.py
+++ b/buildpack/core/runtime.py
@@ -173,6 +173,14 @@ def _activate_license():
   <entry key="license_key" value="{{LICENSE_KEY}}"/>
 </map>"""
 
+    if(strtobool(os.environ.get("MXRUNTIME_License.UseLicenseServer","false"))):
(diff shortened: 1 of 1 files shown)


buildpack/core/runtime.py
```diff
@@ -7,6 +7,7 @@
 import sqlite3
 import subprocess
 import time
+from distutils.util import strtobool
```
jeohist, line 10: `distutils` is deprecated in Python 3.10, could you use the function from [lib.m2ee.util](https://github.com/mendix/cf-mendix-buildpack/blob/eb39e572556ee2802dfa05b8f76226d0bded2fb9/lib/m2ee/util.py#L203-L216)?

bhavinshah-mendix: Thanks @jeohist ! updated as per your suggestion.

jeohist (approved):
LGTM!

PR #784, opened 2024-08-13
MET-2698 - Changing the startup order for user-metering sidecar

rus-kh:
As part of the MET-2698 story, the order of the startup for Users-metering Sidecar was changed to start it after the Runtime is fully started and to avoid blocking of Runtime queries to app's database.


Yornik (approved):
LGTM

PR #849, opened 2025-01-28
Upgrade datadog agents to the latest versions

ekremsekerci:
- Upgrade datadog-cloudfoundry-buildpack version to `4.42.0`
- Upgrade dd-java-agent (trace agent) version to `1.45.2`
- Move on error actions to separate shell scripts
      * We update java opts for `OnError` and `OnOutOfMemoryError` handling
      * New dd agent expect these commands to be in shell scripts (instead of defining them directly)
      * So, we moved those to separate shell scripts

### DIFF
--- a/etc/scripts/on_error.sh
+++ b/etc/scripts/on_error.sh
@@ -1,0 +1,2 @@
+#!/bin/bash
+kill -s USR1 PYTHONPID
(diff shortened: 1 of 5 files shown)


etc/scripts/on_error.sh
```diff
@@ -0,0 +1 @@
+kill -s USR1 PYTHONPID
```
pappachino, line 1: Would be better to add a shebang line

etc/scripts/on_out_of_memory_error.sh
```diff
@@ -0,0 +1 @@
+kill -s USR2 PYTHONPID
```
pappachino, line 1: Would be better to add a shebang line

PR #860, opened 2025-03-13
Update nginx.conf.j2

tieniber:
A customer noted that the sqlite wasm file used for PWAs is not gzipped.

Add application/wasm to gzip_types in nginx.conf.j2


Robinfr:
@tieniber This also requires changes to the mime types: https://github.com/mendix/cf-mendix-buildpack/blob/master/etc/nginx/conf/mime.types since the .wasm file is not registered there.

It would also be good to register the .mjs file under the mime-types since that accompanies the .wasm file, and by registering that, it would also automatically get gzipped.

PR #915, opened 2026-09-03
Feature/sap hana client jar stage

Pri1235:
Added SAP HANA client JAR staging to the CF Mendix buildpack — fetches ngdbc-2.29.11.jar from the Mendix CDN when INCLUDE_SAP_HANA_CLIENT is set.

### DIFF
--- a/README.md
+++ b/README.md
@@ -273,6 +273,16 @@ Selection based on name `hana` and tags `["hana", "database", "relational"]`.
   ],
 ```
 
+#### SAP HANA JDBC Client JAR
+
+To include the SAP HANA JDBC client JAR (`ngdbc.jar`) in your app's classpath at staging time, set the following environment variable:
+
+```shell
+cf set-env <YOUR_APP> MXRUNTIME_IncludeSAPHanaClient true
+```
+
+When enabled, the buildpack fetches the latest `ngdbc.jar` from the Mendix CDN and places it in `model/lib/userlib/`, making it available to the Mendix runtime. If the download fails, staging continues without the JAR and a warning is logged.
+
 ### Connect an External File Store
 
 The Mendix Runtime supports multiple external file stores: AWS S3, Azure Storage and Swift (Bluemix Object Storage).
--- a/buildpack/core/runtime.py
+++ b/buildpack/core/runtime.py
@@ -61,7 +61,7 @@ _SAP_HANA_CLIENT_DEPENDENCY = "sap.hana.client"
 
 
 def _is_sap_hana_client_enabled():
-    return bool(os.environ.get("SAP_HANA_CLIENT_ENABLED", "").strip())
+    return os.environ.get("INCLUDE_SAP_HANA_CLIENT", "").strip().lower() == "true"
 
 
 def stage_hana_client(buildpack_dir, build_dir, cache_dir):
--- a/buildpack/core/runtime.py
+++ b/buildpack/core/runtime.py
@@ -61,7 +61,7 @@ _SAP_HANA_CLIENT_CDN_PREFIX = "/mx-buildpack/sap-hana-client"
 
 
 def _is_sap_hana_client_enabled():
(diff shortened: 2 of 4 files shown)


buildpack/core/runtime.py
```diff
@@ -57,6 +57,44 @@ def is_version_maintained(version):
...
+    return os.environ.get("INCLUDE_SAP_HANA_CLIENT", "").strip().lower() == "true"
+
+
+def _get_hana_jar_name():
+    import requests as req
+
+    version_url = util.get_blobstore_url(f"{_SAP_HANA_CLIENT_CDN_PREFIX}/version.txt")
+    resp = req.get(version_url, timeout=10)
+    resp.raise_for_status()
+    return f"ngdbc-{resp.text.strip()}.jar"
```
mayankmendix: remove the hardcoded string part to line number 84. Update this function as _get_hana_jar_version().
This way string manipulation will be in one place.

buildpack/core/runtime.py
```diff
@@ -57,6 +57,44 @@ def is_version_maintained(version):
...
+def stage_hana_client(build_dir):
+    if not _is_sap_hana_client_enabled():
+        return
+
+    try:
+        dest = os.path.join(build_dir, "model", "lib", "userlib")
+        util.mkdir_p(dest)
+        jar_name = _get_hana_jar_name()
+        jar_url = util.get_blobstore_url(f"{_SAP_HANA_CLIENT_CDN_PREFIX}/{jar_name}")
+        util.download(jar_url, os.path.join(dest, jar_name))
```
mayankmendix, line 85: What if there is no file in CDN? does it throw error?

buildpack/stage.py
```diff
@@ -217,6 +217,7 @@ def cleanup_dependency_cache(cached_dir, dependency_list):
     metering.stage(BUILDPACK_DIR, BUILD_DIR, CACHE_DIR)
     database.stage(BUILDPACK_DIR, BUILD_DIR)
     runtime.stage(BUILDPACK_DIR, BUILD_DIR, CACHE_DIR)
+    runtime.stage_hana_client(BUILD_DIR)
```
mayankmendix: shouldn't stage_hana_client be called from inside stage method, of runtime?

buildpack/core/runtime.py
Pri1235: yes it will raise HTTPError if jar or version.txt file is missing

README.md
```diff
@@ -273,6 +273,16 @@ Selection based on name `hana` and tags `["hana", "database", "relational"]`.
   ],
 ```
 
+#### SAP HANA JDBC Client JAR
```
mayankmendix, line 276: Name is as "SAP HANA Client" only.
Also in the description below.

buildpack/core/runtime.py
```diff
@@ -57,6 +57,44 @@ def is_version_maintained(version):
...
 
 
+_SAP_HANA_CLIENT_CDN_PREFIX = util.BLOBSTORE_BUILDPACK_DEFAULT_PREFIX + "sap-hana-client"
+
+
+def _is_sap_hana_client_enabled():
+    return os.environ.get("MXRUNTIME_IncludeSAPHanaClient", "").strip().lower() == "true"
+
+
+def _get_hana_jar_version():
```
mayankmendix, line 67: We do not need this as a separate method. Can be merged into stage_hana_client

buildpack/core/runtime.py
```diff
@@ -57,6 +57,44 @@ def is_version_maintained(version):
     return False
 
 
+_SAP_HANA_CLIENT_CDN_PREFIX = util.BLOBSTORE_BUILDPACK_DEFAULT_PREFIX + "sap-hana-client"
```
mayankmendix, line 60: Do not need to be a global constant. Can be a local constant for method stage_hana_client

mayankmendix (approved):

PR #916, opened 2026-09-21
Skip staging SAP HANA client JAR if already present in userlib or vendorlib

Pri1235:
Checks both model/lib/userlib and model/lib/vendorlib for an existing ngdbc*.jar before downloading. The Mendix runtime scans both directories, so staging a duplicate is unnecessary.


buildpack/core/runtime.py
```diff
@@ -62,9 +63,22 @@ def _stage_hana_client(build_dir):
...
         return
 
+    model_lib = os.path.join(build_dir, "model", "lib")
+    for lib_dir in ("userlib", "vendorlib"):
+        existing = glob.glob(os.path.join(model_lib, lib_dir, "ngdbc*.jar"))
+        if existing:
+            match = re.search(r"ngdbc-(.+)\.jar", os.path.basename(existing[0]))
+            version = match.group(1) if match else "unknown"
+            logging.info(
+                "SAP HANA client JAR version [%s] already present at [%s], skipping download",
```
mayankmendix: Parsing version is not needed. Just loggin the found file name should be enough.

buildpack/core/runtime.py
```diff
@@ -62,9 +63,22 @@ def _stage_hana_client(build_dir):
     if not enabled:
         return
 
+    model_lib = os.path.join(build_dir, "model", "lib")
+    for lib_dir in ("userlib", "vendorlib"):
+        existing = glob.glob(os.path.join(model_lib, lib_dir, "ngdbc*.jar"))
```
mayankmendix: add a '-' in the search. Make it "ngdbc-*.jar"

mayankmendix (approved):

